PR #18, opened 2014-10-24
Add support for SE orgnr

benregn:
I'm not sure if this is the best way to do this since it's the same as the se/vat.py except for the missing '01' check. Maybe adding extra docs to the vat.py file would be better, e.g. "For checking orgnr append '01' to the number".


arthurdejong:
I've made some modifications and your changes were merged as dd309e4e2279a1c648546f9417c3f624ccbd6f21.

PR #19, opened 2014-12-23
Adding HR (Croatia) country code to the list of available countries.

egcastro:
The module was created but not expose and the HR country code was missing from the list of countries.


arthurdejong:
This was merged as 9883c72176409c6a2cdb8a0f54fe4571446ddcb6.

Sorry it took so long.

egcastro:
Don't worry, thank you for maintain this library.

PR #20, opened 2015-02-09
Add ISO 9363 (BIC) support

tonybajan:


tonybajan:
#21 gets the ISO number right in the branch name, sorry.

PR #22, opened 2015-03-18
Updated imsi.dat

matthewh:
I noticed imsi.dat was out of date. Any chance you could release a new version with this update?

```
python getimsi.py > stdnum/imsi.dat
```


arthurdejong:
The data files are updated as part of a new release (as in 583b066bea38ef0f15ffd17d265d35e4311faac2). The source includes a script for updating the data file. A better mechanism would be to have something that can be run from the installed library.

PR #27, opened 2015-08-03
Estonian Personal ID (isikukood) validation

t0mk:
added validation code plus tests


arthurdejong:
Thanks for your pull request. Your changes have been merged with some modifications. Specifically I simplified the check digit calculation and added some more tests to get 100% code coverage.

PR #28, opened 2015-09-06
add Polish PESEL number

dchoruzy:


arthurdejong:
Thanks for the patch. Can you provide some information in the docstring? Also, can you provide some test cases in the docstring? Ideal would be to have an extra test file (e.g. something like test_br_cnpj.doctest) that contains a number of valid numbers.

arthurdejong:
The PESEL module was implemented in a0cb61f7b1718b93e583e749c1483f0c6a5bdc3a based on your work. Thanks.

PR #29, opened 2015-09-06
add Polish REGON number

dchoruzy:


arthurdejong:
Thanks for the patch. Can you provide some information in the docstring? Also, can you provide some test cases, specifically tests for valid numbers of both 9 and 14 digits? Ideal would be to have an extra test file (e.g. something like test_br_cnpj.doctest) that contains a number of valid numbers.

You can run the tests with nosetests.

arthurdejong:
The REGON module was implemented in 36217ac8d09f0de594b1ba9b17b435329c16548c based on your work. Thanks.

PR #30, opened 2015-10-16
ISO6346 - Bug fix for the case when the checksum is 10

rpaloschi:
When the container code checksum is 10, the function compares it with the last digit of the container code (that, in this case, would be 0).

A simple fix, we just need to compare the last digit in every case.

The fix contains a test that exposes the bug.


arthurdejong:
Thanks for the patch. Your fix was incorporated with a small modification, see fb0efe0e35d4dc3e3fdff87fdf5150ed3368f58d.

PR #34, opened 2016-05-22
French NIR and NIF

DimitriPapadopoulos:
Add module for NIR (French social security number) and NIF (French tax identification number).


arthurdejong:
Hi. Thanks for providing this. I've merged your NIR implementation in a1afa76. I added you as author for that file.

The commit message for NIF mentions NIR and the doctests for that module don't pass. I couldn't find anything about a check digit in the NIF but the there are some limits to what the first digit can be which could be checked for.

arthurdejong:
I have also merged your NIF implementation as 2409ee9. I have added some more comments and an extra reference.

Thanks for your contributions!

PR #35, opened 2016-05-27
add a SIRET module

yaubi:
SIRET stands for "Système d’identification du répertoire des établissements". It's a French company's establishment identification number that appends a 4-digit number to the company's SIREN, plus a single digit for error detection using the Luhn algorithm.

### DIFF
--- a/README
+++ b/README
@@ -51,6 +51,7 @@ Currently this package supports the following formats:
  * HETU (Henkilötunnus, Finnish personal identity code)
  * Y-tunnus (Finnish business identifier)
  * SIREN (a French company identification number)
+ * SIRET (a French company's establishment identification number)
  * n° TVA (taxe sur la valeur ajoutée, French VAT number)
  * SEDOL number (Stock Exchange Daily Official List number)
  * VAT (United Kingdom (and Isle of Man) VAT registration number)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -46,7 +46,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'python-stdnum'
-copyright = u'2013, Arthur de Jong'
+copyright = u'2013, Arthur de Jong. 2016, Yoann Aubineau'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -127,6 +127,7 @@ Available formats
    fi.hetu
    fi.ytunnus
    fr.siren
+   fr.siret
    fr.tva
    gb.sedol
    gb.vat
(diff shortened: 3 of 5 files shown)


arthurdejong:
Thanks for your pull request. A few points for improvements:

- in some places SIREN is still mentioned instead if SIRET
- the link in the comment to the online validation is not relevant for SIRET because it only does SIREN
- please update the copyright header to include your own name
- can you add/update the documentation for the to_siren() and to_tva() functions?
- there is a PEP8 issue (two empty lines before a function definition)

The to_siren() implementation looks a bit ugly and I've been thinking of a way to write it a little more elegantly but haven't come up with something better yet. The join() call can be combined on the return though. It is nice that it preserves formatting, then again I'm not sure that is always a good idea, unless there is some standard way of formatting a SIREN or SIRET.

In the to_siren() function a siren variable is redefined. In the module scope it refers to the stdnum.fr.siren module and in the function it is first a list and later a string.

Another thing that I noticed is that you can probably add siren.validate(number[:9]) to the validate() function (probably after the luhn.validate() call).

Thanks.

yaubi:
Hi Arthur;

Thank you for your feedback. I agree with all of your comments and I will
soon update my commit accordingly.

As for to_siren(), I tried an alternative implementation using
itertools.takewhile() but ended up with even more ugliness. I am open to
any suggestion you could think of.


2016-05-27 23:55 GMT+02:00 Arthur de Jong <notifications@github.com>:

yaubi:
Actually, I am thinking about something like this:

> `>>> import itertools`
> `>>> siret = '987 654 321 00009'`
>
> `>>> it1, it2 = itertools.tee(enumerate(siret))`
> `>>> digits = [(i, c) for i, c in it1 if c.isdigit()]`
> `>>> spaces = [(i, c) for i, c in it2 if c.isspace()]`
> `>>> chars = [c for i, c in sorted(digits[:9] + spaces)]`
> `>>> siren = ''.join(chars).strip()`
>
> `>>> siren`
> `'987 654 321'`

What do you think? It is half the number of lines but I am not sure it is more readable.

arthurdejong:
It is more compact but indeed more difficult to follow. The most compact, readable thing I came up with so far is:

```
def to_siren(number):
    siren = []
    digit_count = 0
    for char in number:
        if digit_count < 9:
            siren.append(char)
            if char.isdigit():
                digit_count += 1
    return ''.join(siren)
```

Something slightly more compact and almost functional is the following but I think I still prefer the imperative version above.

```
import itertools

def to_siren(number):
    counter = itertools.count()
    return ''.join(
        char for char, digit_count in (
            (char, counter.next() if char.isdigit() else 0)
            for char in number)
        if digit_count < 9).strip()
```

yaubi:
Yeah, I prefer the imperative approach as well. Will update the commit early next week.

yaubi:
Commit updated. Is it ok for you?

arthurdejong:
Thanks for your contribution I have merged your commit as 5ba3a87. I've also added some extra tests for numbers found online.

PR #36, opened 2016-05-29
Improve French NIR validation

DimitriPapadopoulos:
Please note that the 7th character of the NIR might be 'A' or 'B'.
Other than that the NIR contains digits only.

### DIFF
--- a/README
+++ b/README
@@ -110,6 +110,8 @@ Currently this package supports the following formats:
  * RTN (Routing transport number)
  * SSN (U.S. Social Security Number)
  * TIN (U.S. Taxpayer Identification Number)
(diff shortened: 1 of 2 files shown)


arthurdejong:
Sorry for not replying sooner. I've merged part of your changes as 879f2d3. I made some further improvements to the module in 43b58d3 including extra tests and some additional documentation and references.

I haven't merged your extra checks for gender, month and COG yet. There seem to be conflicting information on which values they may be. Furthermore it seems that the COG-part of the number can be modified if the serial part would overflow 999.

Feel free to make a separate pull request for that, preferably with some extra tests. For some numbers I also have a test-set of valid numbers that were found online but this is obviously more difficult with personal numbers (unless they are part of some leak).

PR #37, opened 2016-08-26
Support for Spanish CUPS code for energy supply points

vokimon:
CUPS codes identify energy supply points in Spain. They are extensively used on pipeline gas and electricity markets. The format is set by the Energy Ministry, and individual codes are issued by each local distribution company.

### DIFF
--- a/stdnum/es/cups.py
+++ b/stdnum/es/cups.py
@@ -1,7 +1,8 @@
 # cups.py - functions for handling Spanish CUPS code
 # coding: utf-8
 #
-# Copyright (C) 2011-2015 Arthur de Jong
+# Copyright (C) 2016 David García Garzón
+# Copyright (C) 2016 Arthur de Jong
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
--- a/stdnum/es/cups.py
+++ b/stdnum/es/cups.py
@@ -1,0 +1,162 @@
+# cups.py - functions for handling Spanish CUPS code
+# coding: utf-8
+#
+# Copyright (C) 2011-2015 Arthur de Jong
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA
+
+"""CUPS (Código Unificado de Punto de Suministro, Supply Point Unified Code)
+
+CUPS codes are used in Spain as unique identifier for energy supply points.
+They are used both for electricity and pipelined gas.
+
+They have 20 or 22 digits: LL DDDD CCCC CCCC CCCC EE NT
+
+* LL: (letters) Country. (Always 'ES' since it is a national code)
+* DDDD: (numbers) Distribution company code (numeric)
+* CCCC CCCC CCCC: Identifier within the distribution company (numeric)
+* EE: (numbers) check digits
+* N: (number) Border point sequence
+* T: (letter) Kind of border point
+	* 'F': border point
+	* 'P': main metering point
+	* 'R': redundant metering point
+	* 'C': checking metering point
+	* 'X'/'Y'/'Z': registering point
+
+More information:
+
+* https://es.wikipedia.org/wiki/C%C3%B3digo_Unificado_de_Punto_de_Suministro
+
+>>> compact('ES 1234-123456789012-jy')
+'ES1234123456789012JY'
+
+>>> format('ES1234123456789012JY')
+'ES 1234 1234 5678 9012 JY'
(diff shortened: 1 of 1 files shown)


arthurdejong:
Thanks for your patch. I'm working on integrating it. There are some PEP8 issues and consistency with other modules (e.g. use of reference instead of number). I have also just pushed a commit to master (d95382f) that allows using the more specific exceptions with custom messages.

I was also looking into moving some of the tests to a separate doctest file (to not clutter the documentation with tests too much). For a number of other modules I have a list of valid numbers that are in actual use. This ensures that at least the module works with known correct numbers. Do you know if such a list is available somewhere? I usually collect a large number of numbers, test with that and then use a small, representative selection for the actual test.

vokimon:
Sorry for the pep8 issues. :-)

'reference' is not really a good name. Renaming it to 'number' would be an improvement. 'code' would be the semantically accurate name but i understand uniformity is nice for the library.

I don't understand how the changes on ValidatonErrror  enhances flexibility beyond what format does. It just concatenates args but just if they are strings. Notice that some of the args are integers. It is nice because it makes the parameters available to exception handlers, but not in quite straightforward way. Maybe extending ValidationError for each kind of error and expliciting the arguments, but maybe we are over-designing here.

Regarding the tests, I come from TDD tradition and I just added a test for every kind of valid code (in this case, with or without border point) and then a test for every failing condition to be detected. But yes, indeed the condition for the check code is quite a weak test and further modification could generate undetected regressions. So it might be nice to add several succeeding tests with different check codes and their matching failure test changing the check code.

Regarding real CUPS codes I am not allowed to publish real CUPS codes as they are considered sensitive information, and indeed many energy companies seek for them to switch users without their consent (quite illegal but quite impune). On the wikipedia article i linked on the docs, there are several valid CUPS codes although they are not real ones. Check code is not affected by border point digits so you can take the examples and add or remove border points.

-   ES 0987 5432 1098 7654 ZF
-   ES 1234 1234 5678 9012 JY
-   ES 9750 2109 8765 4321 CQ
-   ES 0999 1100 1234 5678 EK

For your peace of mind we are working with 25K of those codes and it validates them, and wikipedia ones as well.

vokimon:
BTW, I don't know whether this goes beyond the scope of the library, but i would like to offer the components of the number with semantic accessors, not just splitting the components but also naming them. For instance for a CUPS you would have:

- distributor_company_id
- supply_id
- point_id
- is_border_point
- is_main_meter_point
- is_redundant_meter_point
- is_registry_meter_point

is there any standard interface in stdnum for that? if so, i will be glad to implement it, but for another PR.

vokimon:
I think i now understand what you mean about the Exceptions. I didn't notice about ValidationError subclasses, and i was puzzled. As i tried to use subclasses instead of the base one, I saw the problem you solved in that commit. Looks perfect, then. Since you are already integrating that i won't but just tell me if you want me to do it. I already applied the changes you say in a third validator I am about to PR.

arthurdejong:
Thanks for comments and the improvements you made to the other pull requests.

The reason for the exception subclasses is so that the caller of the library can handle the exceptions in a nicer way. For example, I didn't want to integrate gettext into python-stdnum but this at least allows some translatability (e.g. a calling library could monkeypatch translatable strings in the exceptions). As such I'm not 100% sure I prefer custom error messages per module.

Regarding the validation. If you can confirm that you ran the check against a large number of valid numbers that is enough for me (that could be better than a few other formats). Thanks.

Regarding the deconstructing the various components of a number: the `stdnum.isbn` module and a few others have a [`split()`](https://github.com/arthurdejong/python-stdnum/blob/master/stdnum/isbn.py#L173) function. Some other modules have a [`get_birth_date()`](https://github.com/arthurdejong/python-stdnum/search?q=get_birth_date) or [`get_gender()`](https://github.com/arthurdejong/python-stdnum/search?q=get_gender) function to get a specific property from a number.

Anyway. I'm almost finished with polishing it. I have one question about the `format()` function:  the function returns a space between the last two digits but none of the examples have that.

vokimon:
Definitely, N and T shoud be together in `format`. I separated them because the are different components and i was thinking in `split()` but most examples display them together, so let's join them in `format`. I must say that CUPS are used mostly in their compact format in electrical invoices and related documentation (and maybe that's why often users fail to type such long code properly).

I fully agree on using exception subclasses and I understand the localization issues. The detailed messages I crafted should be just a help for developers and the application developer should provide their own localized messages for every exception subclass. Maybe, we could let the application developer to parametrize their localized messages if we put attributes in the exception. For example, `InvalidChecksum` could have a `expected_checksum` and `failed_checksum`. But for the rest of exceptions i can't find a general set of attributes.

`split()` does not provide semantics which is the important aspect to me. So providing `get_distributor_company_id()` and others looks a better solution. 

I was thinking on an interface that provides object with attributes (or properties) so that it can play nice in `format`:

For instance:

```python
class components(object):
     def __init__(self, number):
           self.compact=compact(number)
     @property
     def distributor_id(self):
           return self.compact[2:69]
     ....

cups = stdnum.es.cups.components(cups)
iban = stdnum.es.cups.components(iban)
ccc = stdnum.es.cups.components(iban.localpart)
print cups.distributor_id
print "The distributor is {0.distributor_id} and the supply is {0.supply_id} and the bank code {1.bank}".format(cups, ccc)
```
Not for this PR, of course

arthurdejong:
I've merged your change as 466cec8 for now. I've taken out the custom exception messages for now. It makes the code a little more compact and I haven't figured out yet what I want to do with the exceptions.

I've also moved some tests to a separate doctest file and shortened the docstring a bit. I've moved the check digit calculation to separate functions to make the code more readable.

Feel free to open new pull requests for the getting of components from the number or other improvements. I will keep the nicer exceptions around so I can merge those when needed.

Thanks!

PR #38, opened 2016-08-26
Support for Spanish 'Referencia Catastral' for real estate

vokimon:
"Referencia Catastral" (cadastral reference) is the id for real estates in Spain and are issued by the Ministerio de Hacienda (Tresury Department).

Legal documentation is quite obscure about the validation algorithm, so I took as reference an existing javascript implementation (url in comments) that validated every real reference i passed to it.

Also because of that, i was tolerant about the content of the code beyond the control digits. One could deduce from examples in official documentation that references could be further checked for structure (letters, digits, which ones) given the type (urban, non-urban and special reference), but no explicit statement is said about how to identify them. This could be further work but i documented what i found in the module documentation.

### DIFF
--- a/stdnum/es/referenciacatastral.py
+++ b/stdnum/es/referenciacatastral.py
@@ -59,8 +59,8 @@ special (infrastructure) states:
             * 3C: Tolled road
             * 3T: Tolled tunnel
         * Ports and airports
-            * 4A: Aeropuertos:
-            * 4P: Puertos comerciales:
+            * 4A: Airports
+            * 4P: Comercial ports
     * Province (major if many) (2 digits)
     * Municipality (3 digits)
     * Scope (1 letter, leter)
--- a/stdnum/es/referenciacatastral.py
+++ b/stdnum/es/referenciacatastral.py
@@ -94,16 +94,16 @@ More information:
 >>> validate('783301 VG8173B 0001 TT') # Missing digit
 Traceback (most recent call last):
     ...
-ValidationError: Wrong length for cadastral reference \
+InvalidLength: Wrong length for cadastral reference \
 '783301VG8173B0001TT', should be 20
 >>> validate('7837301/VG8173B 0001 TT') # non alphanum
 Traceback (most recent call last):
     ...
-ValidationError: Found invalid digits '/'
+InvalidFormat: Found invalid digits '/'
 >>> validate('7837301 VG8173B 0001 NN') # bad control digits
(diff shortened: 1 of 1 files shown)


vokimon:
I moved forward by addressing the issues you already pointed out in #37

arthurdejong:
Thanks for your patch. I've been working on integrating it. From this page:
https://www1.sedecatastro.gob.es/CYCBienInmueble/OVCBusqueda.aspx
I get the impression that the number can be 7, 14, 18 or 20 digits.

There is also a TODO in the code regarding numbers with 19 digits.

I've found a lot of numbers online and a lot seem to be 14 digits. They sometimes result in a list of numbers when searching via the above form and sometimes result in a similar result to a 20-digit version. I don't speak Spanish so my understanding is all via Google Translate.

vokimon:
No, definitely it is 20 digits long.

19 digits on rustic properties is not true. A typo in some documentation I consulted put me into that doubt for a while. So I wrote the TODO to check the inconsistency. It was clearly a typo since it was a bad copy of http://www.catastro.meh.es/esp/referencia_catastral_1.asp The number on the title for the rustic parcel example was missing the first 1. Then I noticed that the example image had indeed 20 digits, including that missing "1" and then i found the original source just to confirm it. But I forgot to remove the TODO. Sorry for that.

Most other numbers you mention I guess they talk about the number structure. Old numbers (quite old, indeed) had 14, and no check code. Nowadays the official number is a combination of those old 14 identifying the parcel plus 4 identifying the property within the parcel (flats, partitions...) plus the two check digits that are the ones to be validated. Old 14 digit codes do not have check code and it is pointless to check them. Depending on the kind of parcel the parcel number has a different structure. That's why there are a lot of documentation talking about what those 14 digits mean. Their meaning is also explained in English in the documentation I wrote. You can get a legal number referring to what the eighteen-century number meant. If you want to refer the same the old 14 code did, that is, the parcel itself before the regulation, the legal code now is that old code plus (usually) "0000" plus the checkcode.

Regarding the other lengths you mention, 18 digits could refer to the meaningful digits of the code. And 7 digits could refer to the two sets that the 14 parcel split into in urban properties, the map and the parcel within the map. If you point me out the actual documentation you are gtranslating, i could tell you.

arthurdejong:
Thanks for you code and your explanations. I've merged the change as 2c557a0. I've put some of the documentation to the tests file in 49db553.

PR #39, opened 2016-08-31
Support for Spanish "Código Cuenta Corriente" (CCC) for bank accounts

vokimon:
CCC is the Spanish local part of the IBAN.

TODO: "Bank code" component of the number should be checked to match an existing bank.  But the PR does not check that. Documentation provides a link to the data source that would be required to validate it. Spanish Central Bank updates the list every few months and it would require an automated extraction tool to maintain it easily and not aging. Not for this PR. But yet, this PR is quite useful as is.

### DIFF
--- a/stdnum/es/ccc.py
+++ b/stdnum/es/ccc.py
@@ -1,0 +1,137 @@
+# ccc.py - functions for handling Spanish CCC bank account code
+# coding: utf-8
+#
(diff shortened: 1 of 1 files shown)


arthurdejong:
Thanks for providing the module. I've merged it as 294f872.

PR #41, opened 2016-11-08
Implement calc_check_digits in SEPA Identifier of the Creditor (AT-02)

pokoli:


arthurdejong:
Thanks for the pull request. I've merged your commit as a9e5405. There was a bug in the test that I fixed and I've taken the opportunity to also fix a few typos in the module.

PR #44, opened 2016-11-14
Implement CBU (unique bank code) of Argentina

lukio:


arthurdejong:
Thanks for your contribution I've merged your commit as dcde8f4. I simplified the checksum calculation a bit to bring it more in line with the other functions, added some documentation and added a few tests (moving them to a separate doctest).

PR #45, opened 2016-12-01
Add new 47 EIN prefix; fix duplicate 46; move 81

kaiku:
47 appears to be a new Internet campus code. Prefix 46 was listed twice, once under the Philadelphia campus and again under the Internet campus. This error may be seen on the IRS website itself. The Wikipedia article on EIN (https://en.wikipedia.org/wiki/Employer_Identification_Number) does not have 46 listed twice. 81 has moved from the Philadelphia campus to the Internet campus.


kaiku:
Just to clarify: the EIN dat file is now in harmony with the values listed on the EIN wikipedia article referenced above.

arthurdejong:
Thanks for the fix. I've merged it as 7d16ea5 (fix to the commit message). It would be nice to have a clean online source for this information (e.g. similar to some of the other get....py scripts).

kaiku:
Would it be possible to bake a new release (1.6, I think) with this fix?

PR #49, opened 2017-09-04
Add support for German Stuernummer

MuhistheHolvi:
German Stuernummer is an identifier given by the German tax office for companies taxing purpose. The identifier follows a set of rules specified here:
https://de.wikipedia.org/wiki/Steuernummer#.C3.96sterreich
There is still a problem with applying the checksum calculations.


MuhistheHolvi:
I found out today that there is actually two paths for validating the identifiers described here:
http://www.pruefziffernberechnung.de/S/Steuernummer-DE-Berlin.shtml
I will look into them later.

PR #50, opened 2017-09-07
Implement a validator for Austrian Steuernummer

MuhistheHolvi:
Implement a validator for Austrian Steuernummer.

### DIFF
--- a/stdnum/at/stnr.py
+++ b/stdnum/at/stnr.py
@@ -195,7 +195,9 @@ def validate_acceptable_opening_chars(number):
     any tax office"""
     opening_chars = number[:2]
     if opening_chars not in get_all_codes_dict():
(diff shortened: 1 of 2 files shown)


arthurdejong:
Sorry for not merging this sooner. I made the following changes regarding your pr:

- the module is now named `stdnum.at.tin` (instead of `stnr`) because I found that the number is really called Abgabenkontonummer (which is a bit long for code) and the last 7 digits are considered the Steuernummer
- the list of tax offices is now in a separate file
- there are now `info()` and `format()` functions
- I implemented check digit validation as described on https://de.wikipedia.org/wiki/Abgabenkontonummer
- the tests were changed to doctests and a few example numbers were found online to test the validation
Thanks again for providing this code and sorry it took so long to merge.

PR #52, opened 2017-10-02
adjust soap request over checkVatApprox() for ES

waldemar-becker:
In order to have requests with address parameters against european VIES service, and get match values back.
It's reasoned by a special case where VIES behaves different then with oder EU countries.


waldemar-becker:
@arthurdejong  if you have questions regarding this, I am happy to answer them.

arthurdejong:
@waldemar-becker Sorry for taking so long to respond.
Can you update the docstring with a little explanation of the parameters? Do you perhaps have a pointer to explanation of the approx calls?

PR #55, opened 2017-10-14
Close  #54: Ensure string in get_cc_module()

### LINKED ISSUE #54: get_cc_module ensure string from unicode source
Could we assure string over unicode in `get_cc_module()`?
```
cc = u'co'
name = u'vat'
__import__('stdnum.%s' % cc, globals(), locals(), [name])
TypeError: Item in ``from list'' not a string
```

blaggacao:
Closes https://github.com/arthurdejong/python-stdnum/issues/54

### DIFF
--- a/stdnum/util.py
+++ b/stdnum/util.py
@@ -164,7 +164,7 @@ def get_cc_module(cc, name):
     if cc in ('in', 'is', 'if'):
         cc += '_'
     try:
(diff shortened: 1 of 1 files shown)


blaggacao:
Tested during debugging and works as expected

arthurdejong:
Thanks for the fix. I have merged your change as 74c1721 with a minimal change: str() is not needed for the country code argument since the %-interpolation takes care of that. I've also added a few tests for function.

PR #56, opened 2017-11-30
Added India identifiers - Aadhaar + PAN

srikanthlogic:
Added Aadhaar, Indian personal digital ID number.
12 digit number including verheoff checksum digit. Does not start with 0 or 1.
Reference - https://groups.google.com/forum/#!topic/aadhaarauth/eB5hOU-Qtq0

### DIFF
--- a/README
+++ b/README
@@ -9,6 +9,7 @@ Available formats
 
 Currently this package supports the following formats:
 
+ * Aadhaar (Indian digital identity number UID, Indian resident personal identity number)
  * NIPT (Numri i Identifikimit për Personin e Tatueshëm, Albanian VAT number)
  * CBU (Clave Bancaria Uniforme, Argentine bank account number)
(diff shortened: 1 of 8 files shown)


arthurdejong:
Thanks for providing the format!

I've merged it as c576bc4. Can you open a separate PR for the PAN format?

srikanthlogic:
Thanks for merging, submitting PAN as a seperate PR; doctests document files arent required seperately?

arthurdejong:
The changes to README and documentation are made as part of the release process so they are not needed for when adding the number format.
I removed the separate doctests because they were duplicates of the tests that were present in the module docstring. You should be able to run the test suite with nosetests or tox. I usually add a separate docstest file if there are tests that are not really suitable as module documentation.
For instance, for some numbers it is easy to make a list of valid numbers that are found online. These are then all tested to give more confidence in the correctness of the validation. For personal numbers such a list is often more difficult to maken.

PR #57, opened 2017-12-01
Add Indian PAN

srikanthlogic:


arthurdejong:
Thanks for the PR. I'll look into integrating the format.
I did a quick search and found that the PAN does have a check digit but I could not find any details on an algorithm. Do you have any idea what kind of function is used for the check digit? Even if it is something as vague as an ISO validation that would help.
If we have a list of enough valid numbers it could also be possible to reverse-engineer the algorithm but since it is an alphanumeric code it will be a bit tricky.

srikanthlogic:
Thanks much for looking into it.

I am searching hard for checkdigit algorithm myself and yet to find a link. I remember reading it a few years ago somewhere, but its lost on interwebz now. I can provide potentially 1000s of valid PAN, but unsure what can decipher it. May be an ML program(?), I hardly know a thing on ML.  I also have an strange feeling / hunch that someone broke the algorithm when they enabled multiple institutions to be issuers. I have tried luhn mod 36 etc, but without any luck.

I am actually waiting on finding this for my next PR for GSTIN, because 15 digit GSTIN (Goods and Services Tax identifier) encompasses 10 digit PAN, I have GSTIN checksum validator ready, but waiting to find PAN checksum so GSTIN could be doubly validated with dual checksum.

arthurdejong:
Sorry it took so long to merge but I merged it with some modifications the biggest of which is that info() now returns a dict which is more in line with other similar functions in stdnum. I try to avoid using a specific representation format because each application will want to do it in a different way.

srikanthlogic:
Thank you for merging and Happy new year. 

Noted the info changes, shall use same for GSTIN. Hopefully, someday the algorithm for PAN checksum is open.

PR #61, opened 2018-01-07
Please don't rely on stable set ordering.

lamby:
Whilst working on the Reproducible Builds effort [0], we noticed
that python-stdnum could not be built reproducibly as it relies
on a stable set ordering when generating the documentation.

This has been filed in Debian as #88652

 [0] https://reproducible-builds.org/
 [1] https://bugs.debian.org/886522

Signed-off-by: Chris Lamb <chris@chris-lamb.co.uk>

### DIFF
--- a/stdnum/eu/vat.py
+++ b/stdnum/eu/vat.py
@@ -43,7 +43,7 @@ from stdnum.exceptions import *
 from stdnum.util import clean, get_cc_module, get_soap_client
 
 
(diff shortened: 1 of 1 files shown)


arthurdejong:
Thanks for providing the fix. I modified the commit message slightly.
Another option would be to set `PYTHONHASHSEED` when generating the documentation (assuming that it also works for sets).

PR #66, opened 2018-02-28
Validate if 'Individual' part of hetu falls between 900 and 999

ekohalsti:
Add validation for 'individual' part of hetu for when it is between 900 and 999 and pass the checksum validation. This range is reserved for test purposes and is not given to any real person.

### DIFF
--- a/stdnum/fi/hetu.py
+++ b/stdnum/fi/hetu.py
@@ -61,6 +61,15 @@ _hetu_re = re.compile(r'^(?P<day>[0123]\d)(?P<month>[01]\d)(?P<year>\d\d)'
                       r'(?P<control>[0-9ABCDEFHJKLMNPRSTUVWXY])$')
 
 
+class InvalidSSN(ValidationError):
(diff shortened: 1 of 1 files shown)


arthurdejong:
Thanks for your contribution. Do you have a reference to where this is described somewhere?

ekohalsti:
Hello!

It's mentioned for example in wikipedia, but in finnish ofc.

https://fi.wikipedia.org/wiki/Henkil%C3%B6tunnus

"Numeroita 900–999 käytetään tilapäisissä henkilötunnuksissa, esimerkiksi sairaaloissa, kun virallista henkilötunnusta ei tiedetä tai sitä ei ole vielä syntyneelle lapselle annettu. Tilapäisiä tunnuksia ei kirjata väestötietojärjestelmään."

Which translates to:

"Numbers 900-999 are used in temporary personal identification numbers, for example, in hospitals when the official identity code is not known or is not yet issued to a child born. Temporary identifiers are not recorded in the Population Information System."

PR #67, opened 2018-03-19
Add missing letter

pabloadan:
letter 'u' missing in line 62


arthurdejong:
The reason that it is not included is that "für" and "fur" will compare as equal (poor man's fuzzy matching). The list was copied from stdnum.de.stnr. It seems that if anything 'a' and 'o' should also be removed from the list. :wink:

PR #72, opened 2018-05-24
Add iso11649 structured creditor reference

EtcEc:


arthurdejong:
Thanks for your PR. About the naming of the module: I would prefer either stdnum.iso11649 or stdnum.creditor_reference whichever seems more appropriate.

Also, do you by chance have a number of numbers available to run tests against? For a number of other formats there is a list of numbers used to test more corner cases. See for example https://github.com/arthurdejong/python-stdnum/blob/master/tests/test_ar_cuit.doctest.

EtcEc:
Sorry about the delay. My wife decided to give birth! Thanks for the feedback. I followed Your suggestions - have another look.

arthurdejong:
Congratulations and sorry for not merging sooner. Thanks for the PR, I've merged it with only minimal changes.

PR #75, opened 2018-06-18
Improved texts and added links

PanderMusubi:


arthurdejong:
Thanks for the improvements!

PR #88, opened 2018-11-27
Norway Person Number and Sweden Person Number

vihtinsky:
Added support for Norwegian eleven-digit birth number (fødselsnummer) 
and Sweden Person Number


vihtinsky:
I had test with good coverage but travis do not run it - 
https://github.com/vihtinsky/python-stdnum/blob/master/tests/test_no_ssn.py#L44

arthurdejong:
Thank you for your contribution. I've merged the Norwegian number as d11e5c4 and the Swedish one as b77c9cd. I have made some changes to the code (added appropriate copyright headers, changed the tests to doctests and implement all required functions).

Please open a separate PR if you feel the Chinese region validation change is useful. If the forum post mentioned is the only source I would be a bit doubtful about the validity of the numbers (they appear to be generated so that other people can use Chinese identifiers).

vihtinsky:
@arthurdejong 
Thank you. Switched to your original library in our project. 
I monkey patch the Chinese region validation in our code. We will work with China  customer next year.
May will learn more about this Chinese RIC.

P.S. You forgot to add this numbers to README. :)

arthurdejong:
Thanks. Let me know if anything needs to be adapted (or if there are any other Chinese numbers that are useful to validate).
The README is more or less automatically updated on release by the [numlist.py](https://github.com/arthurdejong/python-stdnum/blob/master/update/numlist.py) script based on the docstrings.

PR #98, opened 2019-05-23
Add GTIN (EAN-14) validation

pokoli:
GTIN is like EAN-13 but can accept up to 14 digitis. The check digit algorithm is exactly the same as EAN. 

Here is a wikipedia article explaining it's usage and charcteristics: 

https://en.wikipedia.org/wiki/Global_Trade_Item_Number


arthurdejong:
Thanks for your PR!

Issue #104, opened 2019-06-02
Add validation for New Zealand IRD number


unho:
* https://www.ird.govt.nz/-/media/Project/IR/PDF/2020RWTNRWTSpecificationDocumentv10.pdf (page 13)

Other resources:

* https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/New%20Zealand-TIN.pdf
* https://wiki.scn.sap.com/wiki/display/CRM/New+Zealand

Implementations in another programming languages:
* https://github.com/fnzc/nz-ird-validator
* https://github.com/ognus/ird-nz

PR #112, opened 2019-06-02
Add New Zealand IRD number

unho:
Fixes issue #104.

### DIFF
--- a/README
+++ b/README
@@ -145,6 +145,7 @@ Currently this package supports the following formats:
  * MVA (Merverdiavgift, Norwegian VAT number)
  * Orgnr (Organisasjonsnummer, Norwegian organisation number)
  * New Zealand bank account number
+ * New Zealand IRD number
  * NIP (Numer Identyfikacji Podatkowej, Polish VAT number)
  * PESEL (Polish national identification number)
  * REGON (Rejestr Gospodarki Narodowej, Polish register of economic units)
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -230,6 +230,7 @@ Available formats
    no.mva
    no.orgnr
    nz.bankaccount
+   nz.ird
    pl.nip
    pl.pesel
    pl.regon
--- a/docs/stdnum.nz.ird.rst
+++ b/docs/stdnum.nz.ird.rst
@@ -1,0 +1,5 @@
+stdnum.nz.ird
+=====================
+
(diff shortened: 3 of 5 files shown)


unho:
@arthurdejong Any idea why this is failing?

arthurdejong:
@unho There are a few issues that can be seen [from the build logs](https://travis-ci.org/arthurdejong/python-stdnum/builds/540389675):

- there is not 100% coverage from the tests in the code, this generally means that some tests are missing: if the tests are perhaps useful as documentation you could add it to the module, otherwise you could make separate tests (there appear to be missing tests for the cases where the number contains non-digits and where the number is out of range)
- the rst file references the wrong module, also there is no need to include the changes to the documentation, they are generally all added on release
- you can ignore the flake8 error (my bad) or rebase on current master that should fix that issue

Furthermore, can you look into the following changes:

- Can you extend the module docstring to have a little more background on the number (e.g. is it for generally for identifying persons or companies, etc.).
- It is probably pretty easy to find a number of New Zealand IRD numbers online with a few smart Google searches (e.g. search for company contact pages with `site:nz`). It would be really nice to have a dozen or so extra numbers that can be checked.
- Regarding the tests for failed numbers in the doctest file: it is probably a good idea to just call them at the top to also show how they fail (e.g. see https://github.com/arthurdejong/python-stdnum/blob/master/tests/test_ve_rif.doctest#L29)
- Can you move the InvalidLength test from the module to the doctest because that would make the module documentation a little shorter.
- Can you change `calc_check_digit()` to return a string instead of an integer to match the other implementations?
- In general other python-stdnum functions try to avoid blank lines inside functions.
- The 
   ```
   remainder = total % 11
   return 0 if remainder == 0 else 11 - remainder
   ```
   bit in the `_calculate_digit()` function can be replaced by
   ```
   return -total % 11
   ```
   (or even integrated into the `sum()` line to make it a single expression.

unho:
@arthurdejong I've made several of the suggested changes and Travis now passes. I left out removing blank lines as IMHO that seriously compromises readability. I also did not look for more examples to test. Regarding adding more information in the docs I am not sure which information to put there.

arthurdejong:
I've merged your PR as 7211ccb with a few changes.

unho:
Thanks for merging the changes, despite I still think that those changes compromised code readability.

There are two other PRs wanting to be landed :-)

PR #123, opened 2019-06-09
Add Paraguay RUC number validation

unho:


unho:
Rebased PR.

arthurdejong:
Thanks. I've squashed the changes into one commit and reduced the test set to try to find a balance between completeness of the test suite and the time it takes.

unho:
Thanks. Will try not to add so much tests the next times.

arthurdejong:
No problem, I often add add numbers I found while developing the module and only strip the tests to around 200 numbers when I'm done. I do try to ensure that the chosen numbers cover most kind of numbers (e.g. all check digit values and number types). For this I used shuf to pick 150 numbers from the full list, and 20 from each category, then took 200 from that list.

PR #129, opened 2019-06-10
Add South Africa TIN number validation

### LINKED ISSUE #108: Add support for South Africa TIN
* https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/South-Africa-TIN.pdf
* Algorithm explained in Appendix D in https://c.ymcdn.com/sites/sait.site-ym.com/resource/resmgr/2014_SARS/Draft_SARS_External_BRS_2014.pdf
* Algorithm also explained in Appendix J in https://old.juta.co.za/law/media/filestore/2012/11/1_Updated_BRS_on_Tax_Dividends.pdf
* According to description in page 41 of https://c.ymcdn.com/sites/sait.site-ym.com/resource/resmgr/2014_SARS/Draft_SARS_External_BRS_2014.pdf the "Tax Reference Number" has a length of 10 and is fully composed of digits, and can can only start with 0, 1, 2, 3 or 9.

unho:
Fixes #108.

### DIFF
--- a/stdnum/za/__init__.py
+++ b/stdnum/za/__init__.py
@@ -1,0 +1,21 @@
+# __init__.py - collection of South Africa numbers
+# coding: utf-8
+#
+# Copyright (C) 2019 Leandro Regueiro
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


unho:
Rebased PR.

arthurdejong:
Thanks. I've merged it as 1e814ce. The check digit was simply the Luhn algorithm so I switched it to use that implementation.

Also I removed the code that strips the "ZA" prefix from the number because the number does not have that prefix normally. There are a few numbers that do have this (the EU VAT numbers are commonly written with the country code prefix and some have even switched to using that by default in communication).

unho:
I was worried about stripping out the "ZA" prefix because Odoo always stores TIN numbers with the country code prefix, but I've just verified that the country code is stripped out before checking it: https://github.com/maestrano/odoo/blob/master/addons/base_vat/base_vat.py#L128-L129

PR #151, opened 2019-07-31
Add string length constrain for iso6346 calc_check_digit function

lukaszgolanski:
I am very excited to make my first GitHub contribution, please be lenient if I have made something wrong in the process.
As a person active in the container leasing business, I was happy to see your package. It is very good and I am using it not only for iso6346 numbers verification but also for Polish VAT No verification (NIP number).
My fix puts constrain on length of the calc_check_digit being checked.
Without it, there is a risk for hidden errors to occur in the calculation.
I have used the same syntax as found in the 'validate' function.

### DIFF
--- a/stdnum/iso6346.py
+++ b/stdnum/iso6346.py
@@ -61,6 +61,8 @@ def calc_check_digit(number):
     """Calculate check digit and return it for the 10 digit owner code and
     serial number."""
     number = compact(number)
+    if len(number) != 10:
+        raise InvalidLength()    
     alphabet = '0123456789A BCDEFGHIJK LMNOPQRSTU VWXYZ'
     return str(sum(
         alphabet.index(n) * pow(2, i)


arthurdejong:
Hi Lukasz,

Thanks for your contribution and sorry for not responding sooner.

If you are using the `calc_check_digit()` function for generating valid container numbers I recommend calling `validate()` after generating a digit to ensure the other aspects of the number are also correct (such as length, format and construction). In general, the check digit calculation functions only calculate a check digit and do not perform any other validations.

Another problem is that some check digit functions expect the whole number to be passed, some can handle both complete and incomplete numbers. As such there is not really a standard way to construct a valid number for most formats.

I will give this some more thought on this. One question also is what kind of validation would be appropriate for the check digit functions (e.g. length, formatting, components).

arthurdejong:
Closing this PR for now due to inactivity, feel free to re-open if this is still relevant.

PR #153, opened 2019-08-30
Fixing New Broke Link of DGII Page

jeffryjdelarosa:
DGII has changed its page, all the link have been broken, so they changed some URL.


arthurdejong:
Thanks!

PR #159, opened 2019-10-01
Add personalid alias for DK and IS

ahaeber:
Add missing personalid aliases for DK and IS to make them work together with get_cc_module.


arthurdejong:
Hi Andreas,

Thanks for providing these aliases (there are probably also other modules that could use aliases). I've merged your changes as e5d7d9f and d43ad77.

PR #162, opened 2019-10-11
Add South Korean Resident Registration Numbers

DimitriPapadopoulos:


DimitriPapadopoulos:
I just need to add tests, right? Hadn't worked on this for a long time, perhaps these coverage tests are something new?

arthurdejong:
Hi Dimitri,

Thanks for the contribution. Yes the tests are the most important missing thing. You can run them locally with tox (use `tox -e py27` to only run the Python 2.7 tests).

Ideally I would like a few small tests in the docstring and more complete tests in a separate doctest file that bring the coverage to 100%. I usually include about 100 valid numbers in the tests that I find online but this is sadly not always possible (especially for numbers that identify individuals).

DimitriPapadopoulos:
Added some tests but unable to reach 100% - and I don't understand why.

arthurdejong:
Hi Dimitri,

Thanks for providing the tests. I've made a few minor changes and merged it as 790a052.

There were only two bits of missing coverage: the number you used with the invalid birth place also had an invalid birth date and the format() function was never called with a number that did not have  13 digits.

DimitriPapadopoulos:
It looks like the documentation (or the code) is wrong:
> If allow_future is True birth dates in the future are rejected.

It should be:
> If allow_future is False birth dates in the future are rejected.

arthurdejong:
Thanks. Fixed in 54218b1.

PR #164, opened 2019-10-11
Change DGII URL and Validation

jeffryjdelarosa:
DGII has changed their URL to validate, so with first take by BeautifulSoup __EVENTVALIDATION and __VIEWSTATE and then apply to make the post and the validation.


jeffryjdelarosa:
making other change

PR #166, opened 2019-10-13
AT: Provide personalid as an alias from vnr

ahaeber:


arthurdejong:
Thanks, merged as 6119590.

PR #167, opened 2019-10-13
AU: Provide vat as alias from tfn

ahaeber:


arthurdejong:
Thanks, merged as 63a643f.

PR #168, opened 2019-10-14
We discard the RNN if allow_future is False

DimitriPapadopoulos:


arthurdejong:
Thanks for the PR. THis was fixed in 54218b1.

PR #170, opened 2019-11-02
Swiss ESR verification

kurtkeller:
Added a module for Swiss ESR numbers (reference number on Swiss payment slips).

- With the code in stdnum/ch/esr.py I'm confident. However, I don't know whether another module does compatible verification. I created the verification from scratch using the available online docs from the Swiss payment services provider (SIX) and PostFinance (financial institution).
- With the files docs/stdnum.ch.esr.rst and tests/test_ch_esr.doctest I'm not really sure what they do; I tried to create them by looking at similar files.
- The changes to ChangeLog, README and docs/index.rst are in a separate commit, because I don't know whether you want to do that by yourself to keep things consistent.

I added this module because I started to use qrbill for creating the new standard Swiss payment slips. qrbill uses validators for IBAN validation, but sofar does not do any validation on reference numbers. I intend to change validations over from validators to stdnum, since stdnum has all the required validations, excpet for ESR, which can be added with this pull request.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2019-11-02  Kurt Keller <Kurt@pinboard.jp>
+
+	* stdnum/ch/esr.py,
+	  tests/test_ch_esr.doctest, docs/stdnum.ch.esr.rst: Add Swiss
+	  ESR numbers
+
+	  The Swiss ESR/ISR/QR-reference is a reference number used on
+	  payment slips.
+
 2019-10-27  Arthur de Jong <arthur@arthurdejong.org>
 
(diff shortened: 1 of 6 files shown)


arthurdejong:
Hi Kurt,

Thanks for your contribution. I've merged it as a45d4f7. The changes to README, ChangeLog and documentation are more or less automatically done on release.

I've only made a few smaller changes to the code. The `compact()` function now strips leading zeros to ensure that all equivalent numbers end up as the same value after `compact()`. I've also rewritten the `calc_check_digit()` function to be iterative instead of recursive and made the implementation more compact.

PR #171, opened 2019-11-04
add format to iso11649

kurtkeller:
add format to iso11649

The iso11649 module does support format(); add it.

According to info from https://www.paymentstandards.ch
and also the wikipedia page already referenced in the
module, the format is whitespace separated groups of
4 characters, where the last group can be less than
4 characters.


arthurdejong:
Hi Kurt,

I've merged your change as 388bac9. Thanks for providing it!

PR #172, opened 2019-11-12
Fix typos

jwilk:
Brazil**l**ian → Brazilian


arthurdejong:
Thanks for the fix. I've merged it as 831c669 (I took out the ChangeLog entry because it is automatically updated on release).

Issue #173, opened 2019-11-22
SCHLUMBERGER NV ISIN 'AN8068571086' invalid


nocluebutalotofit:
![SCHLUMBERGER NV](https://user-images.githubusercontent.com/45629760/69433259-927b3100-0d3b-11ea-9e83-dd3ab9b100cb.png)

isin.is_valid('AN8068571086')
Out[5]: False

PR #174, opened 2019-11-22
Update isin.py

nocluebutalotofit:
added 'AN' to _iso_3116_1_country_codes so that the SCHLUMBERGER NV ISIN 'AN8068571086' is valid.

Fixing issue SCHLUMBERGER NV ISIN 'AN8068571086' invalid #173


arthurdejong:
Hi @nocluebutalotofit,

Thanks for your patch, I've merged it along with the changes in #176 in 87c195f.

PR #178, opened 2019-12-19
This pull request is to fix a warning messages while using do/ncf.py

herohat:
Getting rid of the warning: No parser was explicitly specified, so I'…m using the best available HTML parser for this system ("lxml")


arthurdejong:
The biggest problem with this is that I think the old version of BeatifulSoup does not support this option and will result in a failure.
In de50109 I've just switched from BeatifulSoup to lxml directly for the HTML parsing which should get rid of this problem.
Thanks for reporting this.

Issue #179, opened 2019-12-29
incorrect Argentinian CUIT .format()


eamanu:
Hi,

This issue was reported by @eduzen on [1].

I am not a expert but reading [2] and this not official webpage [3],
the CUIT is composed for [2 digits]-[8 digits]-[1 digit]: ##-########-#.

And a CUIT just accept numbers (and 2 hyphen).

Also, the first 2 digits just accept some numbers listed on [3] (20, 27, 23, 24, 25, 26)
and some others for international purpose (50, 51, 55).

I can work on a patch for this issue if there are not oppositions :)

[1] https://github.com/PyAr/asoc_members/pull/150#discussion_r361808212
[2] https://es.wikipedia.org/wiki/Clave_%C3%9Anica_de_Identificaci%C3%B3n_Tributaria
[3] https://www.serautonomo.net/argentina/que-es-el-cuit.html

PR #181, opened 2020-01-06
Add regex to validate the correct format of CUIT

eamanu:
According to #179 not any (2 digits) prefix are accepted on
argentinian CUIT. So I added an Regex to validate
correctly the CUIT's format.


arthurdejong:
Hi @eamanu,

Thanks for your patch. I've only added the checking of the first two digits and not the checking for the place of the dashes. These kind of separators are generally ignored in python-stdnum because for a lot of numbers their placement is not very well defined.

I've merged your change as 0b30c4b and also ensured that the correct exception is raised (InvalidComponent).

eamanu:
@arthurdejong thanks!

PR #183, opened 2020-01-07
[FIX] nl: dutch vat check broken since 01/01/2020

jhk-odoo:
New vat numbers received by companies after the 1st january 2020
are considered invalid.

### DIFF
--- a/stdnum/nl/bsn.py
+++ b/stdnum/nl/bsn.py
@@ -47,6 +47,36 @@ InvalidLength: ...
 from stdnum.exceptions import *
 from stdnum.util import clean, isdigits
 
+char_to_int = {
+    'A' : 10,
+    'B' : 11,
+    'C' : 12,
+    'D' : 13,
+    'E' : 14,
+    'F' : 15,
+    'G' : 16,
+    'H' : 17,
+    'I' : 18,
+    'J' : 19,
+    'K' : 20,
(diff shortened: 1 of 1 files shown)


jhk-odoo:
@arthurdejong let me know what you think about this

arthurdejong:
Can you provide of this change to the check algorithm? Where do the letters come from?

Also, the tests fail due to an undefined `check_val_sole` and these changes to BSN will not fix BTW because the format is checked which disallows non-numeric digits before the checksum algorithm is applied.

nim-odoo:
Another VAT number: NL002455799B11
FYI: https://www.belastingdienst.nl/wps/wcm/connect/bldcontenten/belastingdienst/business/vat/new-vat-id/

timmuller:
> Fixes the issue for me, tested with vatnumber: NL000099998B57
> 
> vatnumber from: https://www.dezaak.nl/467/zo-ziet-btw-nummer-eruit.htm

How can this work, it contains a syntax error, causing all tests to fail?

nim-odoo:
@timmuller The PR is not correct. I'll adapt what we have done in https://github.com/odoo/odoo/pull/42681

arthurdejong:
Thanks everyone for helping fix this. I've implemented the fix in a9b3e90. I will make a new release quickly.

PR #184, opened 2020-01-07
removed bsn validation in dutch vat number validate function

### LINKED ISSUE #182: New dutch VAT numbers should no longer always confirm to bsn format
The first 9 digits of the vat number where validated for a check digit, like the bsn number that was part of it. However, this is no longer necessarily the case since 01-01-2020 (because of the privacy of self employed persons). Therefore this check should be removed.

Schrooms:
fixes #182


Schrooms:
closed in favor of #185

PR #186, opened 2020-01-22
do: Add missing NCF document types for validation

gustavovalverde:
This document types are in the official documentation, but were not being validated: https://dgii.gov.do/cicloContribuyente/facturacion/comprobantesFiscales/Paginas/tiposComprobantes.aspx


arthurdejong:
Hi Gustavo, thanks for providing the fix. It was merged as 77a77ba.

PR #187, opened 2020-01-31
Add get_birth_date to no.fodselsnummer

lemonsaurus:
This PR adds the ability to get the birth date from a Norwegian fødselsnummer, and also makes minor changes to the Swedish and Danish personal identity numbers (which share many similarities) in order to improve consistency between these three files.

This has the side effect of making validation of all three of these personal identity numbers stricter than it was before, because the validator will now check that the birth date is valid. This appears to be a fairly established convention in the repository already, and makes the validator far more precise.

## Changes
- Improved an outdated docstring in `se.personnummer.get_birth_date`
- Implemented `no.fodselsnummer.get_birth_date`
- Changed the `no.fodselsnummer.validate` validator to account for birth dates.
- ~~Added a check in `se.personnummer.validate` to make sure the birth date is not a future date.~~
- Added a check in `dk.cpr.validate` to make sure the birth date is not a future date. This resolved a TODO from this file.
- Added attribution according to established norm in the rest of the repo.
- Added documentation with examples of new functionality, and updated some documentation (the validators) that was no longer accurate.

### DIFF
--- a/stdnum/dk/cpr.py
+++ b/stdnum/dk/cpr.py
@@ -1,6 +1,7 @@
 # cpr.py - functions for handling Danish CPR numbers
 # coding: utf-8
 #
+# Copyright (C) 2020 Leon Sandøy
 # Copyright (C) 2012-2019 Arthur de Jong
 #
 # This library is free software; you can redistribute it and/or
--- a/stdnum/dk/cpr.py
+++ b/stdnum/dk/cpr.py
@@ -1,4 +1,5 @@
 # cpr.py - functions for handling Danish CPR numbers
+# coding: utf-8
 #
 # Copyright (C) 2012-2019 Arthur de Jong
 #
@@ -44,10 +45,12 @@ More information:
 Traceback (most recent call last):
     ...
 InvalidComponent: ...
+>>> validate("2110525629")
+Traceback (most recent call last):
+  ...
+InvalidComponent: The birth date information is valid, but this person has not been born yet.
 >>> get_birth_date('2110620629')
 datetime.date(1962, 10, 21)
->>> get_birth_date('2110525629')
-datetime.date(2052, 10, 21)
 >>> format('2110625629')
 '211062-5629'
 """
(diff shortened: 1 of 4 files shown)


arthurdejong:
Hi Leon,
Thanks for providing this PR. Here are a few points for improvement:

- Have a look at [the seven rules of a great Git commit message](https://chris.beams.io/posts/git-commit/#seven-rules)
- Please ensure the tests pass:
  - There is a missing file encoding in `stdnum/dk/cpr.py` which is required for Python 2
  - The exceptions now have a text body that ensures the test response is different
  - Some behaviour changes were made where previously valid numbers are no longer valid: please update the tests
  - Please write new tests to ensure the 100% coverage goal is still met
  - The flake8 tests (code style) checks currently fail
- Please combine fixup commits into the original commits (e.g. 1adf61a and 08140f1 can probably be rolled into add8dca and the changes in 3604fc6 should probably be spread over other commits)
- Can you explain why the numbers in `test_no_fodselsnummer.doctest` that were found online (they should have been real numbers at the time) are no longer valid?

lemonsaurus:
> Can you explain why the numbers in test_no_fodselsnummer.doctest that were found online (they should have been real numbers at the time) are no longer valid?

They cannot have been real numbers. The dates from these numbers were invalid. That is to say, they could have been otherwise valid numbers with impossible birthdates, but such a fødselsnummer would never be assigned a person in Norway. Technically valid isn't as useful as valid.

They did not fail previously because the `is_valid` check did not check birthdates until this PR.

I'll make the requested changes ASAP, didn't realize I needed to ensure compatibility with Python 2, my bad.

lemonsaurus:
Oops, my apologies for the multiple force pushes messing up this PR, thought I had a branch checked out.

lemonsaurus:
@arthurdejong I think I've addressed your feedback now. If I've forgotten or misunderstood something, let me know and I'll get it sorted out.

arthurdejong:
Hi @lemonsaurus,
I've merged your changes as f5e0456, 4500881 and df9f922. I only made some minor formatting changes and I've combined a few commits.

Thanks!

PR #190, opened 2020-02-18
[FIX] do: Change DGII endpoint to new one

gustavovalverde:
Old endpoint has been deprecated


arthurdejong:
Hi Gustavo,
Thanks for the fix. I also fixed the response parsing to try to pick the correct properties and merged it as efa70f1.
It is a little troubling that this interface changes so often though...

PR #191, opened 2020-02-24
Optional parameter for allowing temporary range individual numbers

lekksi:
There are valid use cases for the temporary range individual numbers - this PR adds validating  temporary range individual numbers possible with an optional parameter.

The validate function raises `InvalidComponent` for temporary range individual numbers by default as before.


arthurdejong:
Thanks for the improvement. I've merged it as 353db92.

Issue #101, opened 2019-05-27
Add support for South Korea TIN


unho:
* It is called **사업자 등록 번호** in Korean.
* http://world.salestaxhandbook.com/kr-south-korea
* https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/Korea-TIN.pdf
* https://wiki.scn.sap.com/wiki/display/CRM/South+Korea
* http://world.salestaxhandbook.com/kr-korea

Online search:
* http://www.ftc.go.kr/www/bizCommList.do?key=232

* Examples:
  * List with plenty of examples: https://www.psys.co.kr/index.php?pname=myvender&type=wide
  * I was unable to find any list of numbers online, but the number is specified in some certificate, and searching for images of such certificate (using its korean name) we can extract a bunch of examples: https://www.google.com/search?q=%EC%82%AC%EC%97%85%EC%9E%90%EB%93%B1%EB%A1%9D%EC%A6%9D&tbm=isch
    > 214-87-98889
617-86-11575
114-86-00579
221-83-01195
220-87-57205
116-82-00276
615-82-11718
220-82-05060
129-81-16573
119-86-79111
134-86-72683
206-18-12345
148-05-00404
209-82-06382
124-86-85333
383-37-00143
321-81-00982
113-02-43949
113-86-21886
133-81-32077
220-87-18070
116-82-00276
128-81-63302
375-10-00859
311-81-22413
305-81-38564
193-86-00010
220-81-66148
221-81-32210
113-85-26486
312-12-25168
119-87-08063
156-23-00883
649-82-00117
503-86-09407
122-81-62455
504-81-85233
514-81-68599
130-81-92528
680-87-00210
279-81-01263
134-86-24634
409-86-33769
135-81-06333
133-81-40621
134-86-44004
615-86-09068
374-81-01516
106-82-07200
231-81-02896
128-39-49844
682-85-00885
639-86-00985
183-98-00113
119-02-43947
133-81-29441
415-06-83258
114-86-61464
130-86-79710
110-12-39727
533-85-00526
135-09-24640
301-82-62166
766-88-00563
621-15-41743
515-81-10315
214-87-78980
609-81-99124
303-82-07378
109-81-33637
314-86-53230
110-14-16596
138-81-44092
141-81-18585
219-01-39810
308-81-41001
616-24-46281
130-86-89294
201-81-54845
884-17-01026
156-23-00883
202-81-42420
760-06-00079
[...]

PR #197, opened 2020-03-07
Add support for South Korea TIN Number

unho:
Sadly I have not been able to find the algorithm to calculate the check digit. I sent an email to the contact point, but I still have not got any response. Anyway this provides basic format checking that should be of some help.

Part of #101


unho:
@arthurdejong I don't see what is breaking the tests, any idea?

unho:
@arthurdejong I have found the cause.

unho:
@arthurdejong IMHO this can be landed with check digit validation. If we manage to find out how the algorithm is we can add it later.

arthurdejong:
Thanks, I've merged it as b3891f1 without the check digit validation for now. I've also added a function to perform an online check in 1b7e985.

PR #227, opened 2020-06-01
Introduce validator for GB UTR

MuhistheHolvi:
Introduce validator for GB UTR.

### DIFF
--- a/stdnum/gb/utr.py
+++ b/stdnum/gb/utr.py
@@ -54,6 +54,7 @@ def validate(number):
         raise InvalidFormat()
     if len(number) != 9:
         raise InvalidLength()
+    return number
 
 
 def is_valid(number):
--- a/stdnum/gb/utr.py
+++ b/stdnum/gb/utr.py
@@ -29,8 +29,8 @@ More information:
 
 * https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/UK-TIN.pdf
 
->>> validate('9999999999')
-'9999999999'
+>>> validate('123456789')
+'999999999'
 >>> validate('12345678')
 Traceback (most recent call last):
     ...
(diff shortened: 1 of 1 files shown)


MuhistheHolvi:
I need to learn using doctest locally. This constant pushing is not sustainable after every couple lines of change.

MuhistheHolvi:
@arthurdejong Thank you for the pointer. I think I will close this PR when done and open new one with one commit. I think I am approaching one commit per line of code :)

arthurdejong:
HI @MuhistheHolvi,

Sorry for not responding sooner.

You should be able to run the tests locally with tox (e.g. use `tox -e py36` to run it with a Python 3.6 interpreter, and use `tox -e flake8` to run the style tests).
There are still a few issues:

- it only accepts numbers with 9 digits in length while it the documentation (and other sources) claim the number is 10 digits
- it does not strip commonly a start or end 'k' from the number as described on https://design.tax.service.gov.uk/hmrc-design-patterns/unique-taxpayer-reference/
- while an algorithm for the UTR is not well-documented, an explicitly denied FOI request did publish the algorithm via an online checker: https://www.whatdotheyknow.com/request/unique_tax_reference_number (algorithm also found in a few other places)

A function to calculate a check digit based on the above would be (note that the check digit is the first digit, not the last one which is more common):

    def calc_check_digit(number):
        """Calculate the check digit for the number. The passed number should not
        have the check digit (the first one) included."""
        weights = (6, 7, 8, 9, 10, 5, 4, 3, 2)
        return '21987654321'[sum(int(n) * w for n, w in zip(number, weights)) % 11]

The above implementation also seems to match the one used on https://easydigitalfiling.com/kb/what-is-a-utr-number

These generated numbers are treated as correct on both implementations:

    1234567895
    1478641088
    1955839661
    2172858530
    2234567890
    4748949890
    5179265754
    5816619590
    8258106771
    9624088948

PR #228, opened 2020-06-06
Foreign with Ced on Ecuador

cormaza:
For foreign people with ced in Ecuador add support to number 6 in third place


arthurdejong:
Thanks for your contribution, it was merged as 127fff1. I noticed that you referenced the number as "Ced" and not "CI". Is that a more common abbreviation for this number?

cormaza:
Yes, Ced if "Cedula de Identidad" sorry for my english, thanks a lot

Issue #232, opened 2020-08-01
Dominican RNC - escape new line character


evaldez89:
Current situation:  the request crashes when consulting a RNC which result contains a new line character (`\n`) on the name.

The code is already scaping the tab character (`\t`).

When debugging making the request with the RNC 132070801, we can see the response, the issue is when converting the result:
`~stdnum/do/rnc.py(147)`
![image](https://user-images.githubusercontent.com/12959163/89091205-bf292300-d375-11ea-81fc-2c271c6d0899.png)

It crashes when trying to `_convert_result`
`~stdnum/do/rnc.py(115)`
![image](https://user-images.githubusercontent.com/12959163/89091247-1a5b1580-d376-11ea-9c00-1898d1c9812b.png)

I propose scaping the `\n` replacing it with `\\n`

```python
for key, value in json.loads(result.replace('\n', '\\n').replace('\t', '\\t')).items())
```

arthurdejong:
Thanks for reporting this. It is fixed in feca1fe (I've also added a test for this situation).

PR #233, opened 2020-08-01
Escape new line character #232

evaldez89:


evaldez89:
What's do you think about escaping all backslash `\` with a double backslash `\\` ?

PR #234, opened 2020-08-02
Create onrc.py

dotbit1:
First time github, python user. I hope I did everything right.
Please do test!

### DIFF
--- a/stdnum/ro/onrc.py
+++ b/stdnum/ro/onrc.py
@@ -1,0 +1,78 @@
+# onrc.py - functions for handling Romanian ONRC numbers
+# coding: utf-8
+#
+# Copyright (C) 2020 Dimitrios Josef Moustos
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


arthurdejong:
I'm working on merging this with a few extra checks. Thanks for providing the details.

Btw, the build failure is mostly due to missing tests. A number of lines are not covered by the tests. While the output indicates which lines are not touched by testing the easiest way to find out which lines are not executed is to run the tests yourself (e.g. `tox -e py37` to test with Python 3.7) and look at `coverage/index.html`.

stdnum/ro/onrc.py
```diff
@@ -0,0 +1,78 @@
...
+Traceback (most recent call last):
+    ...
+>>> validate('J52/100000/2012') # second number too large
+    ...
+>>> validate('J52/750/3000') # last number (year) is in the future
+Traceback (most recent call last):"""
+
+# regular expression to check ONRC structure
+_onrc_re = re.compile(r'^[JFC](?P<county>[1-9]\d?)\/(?P<n>[1-9]\d{0,4})\/(?P<year>\d{4})$')
+_current_year=datetime.datetime.now().year
```
arthurdejong, line 48: Doing it this way means that the current year is stored on the module level which may cause problems for longer-running applications. Let's say you started the application that uses this code in December 2020 then in January 2021 numbers for 2021 would be considered invalid.

dotbit1: Thank you for considering this case. My intention was speeding up by not asking for year every time, but you are right. Thank you for adding onrc number validation.

Issue #113, opened 2019-06-02
Add support for Malaysia TIN


unho:
* http://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/Malaysia-TIN.pdf
* https://wiki.scn.sap.com/wiki/display/CRM/Malaysia

PR #237, opened 2020-09-10
Add malaysian itn number

pokoli:
This PR implements what is described on #113

I did not found so much information of the number so it is based on:

http://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/Malaysia-TIN.pdf

I will complete the tests with all of the numbers of this source document plus a real number example which I am waiting to receive from some of our users.

### DIFF
--- a/stdnum/my/itn.py
+++ b/stdnum/my/itn.py
@@ -1,0 +1,91 @@
+# itn.py - functions for handling ITN numbers
+#
+# Copyright (C) 2020 Sergi Almacellas Abellana
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA
+
+"""ITN No. (Malaysian Income Tax Number)
+
+The number is assigned by The Inland Revenue Board of Malaysia (IRBM) and it
+is required to report the income. This unique number is known as
+"Nombor CukaiPendapatan" or Income Tax Number.
+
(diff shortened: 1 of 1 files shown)


stdnum/my/itn.py
```diff
@@ -0,0 +1,91 @@
...
+    if len(number) > 12:
+        index += 11
+    return number[:-index], number[-index:]
+
+
+def validate(number):
+    """Check if the number is a valid NRIC number. This checks the length,
+    formatting and birth date and place."""
+    number = compact(number)
+    if len(number) > 13 or len(number) <= 10:
```
arthurdejong, line 71: This also accepts numbers of length 13. Probably use:

    if len(number) < 11 or len(number) > 12:

or

    if len(number) not in (11, 12):

stdnum/my/itn.py
```diff
@@ -0,0 +1,91 @@
...
+    """Convert the number to the minimal representation. This strips the
+    number of any valid separators and removes surrounding whitespace."""
+    return clean(number, ' -*').strip()
+
+
+def split(number):
+    number = compact(number)
+    index = 10
+    if len(number) > 12:
+        index += 11
```
arthurdejong, line 63: Why is this needed if numbers of length greater than 12 are not allowed?

arthurdejong:
Hi @pokoli,

Thanks for looking into this.

The OECD document is pretty unclear but regarding the length it talks about maximum lengths so the actual numbers may be shorter (e.g. "SG 587682020"). The maximum total length appears to be 13 (e.g. "SG 10234567090" is valid).

It may be better to not split on length but split on non-digit and digit parts and check the parts.

These appear to also be links with useful information:

* http://www.hasil.gov.my/bt_goindex.php?bt_kump=2&bt_skum=4&bt_posi=1&bt_unit=4&bt_sequ=1
* http://edaftar.hasil.gov.my/info.php
* http://edaftar.hasil.gov.my/infoc.php

It would be really nice if we could find a list of number types somewhere (the document has a table with categories that mention SG, OG, C, CS, D, E, F, FA, PT, TA, TC, TN, TR, TP, TJ annd LE but in other places TF is also mentioned).

pokoli:
I've got some feedback from our users that recieved a number `552815-W` of company in Malaysia but the number does not match any of the specified numbers. I've checked on the net and the company exists. 

This is strange

arthurdejong:
It cloud be that "552815-W" is a registration number (perhaps with something like a chamber of commerce) and not a tax number. It would really be helpful to have good documentation on Malaysian numbers or someone with local knowledge.

unho:
I don't see any tests.

PR #240, opened 2020-10-02
Expanded validation for BE VAT numbers

JonasGe:
Specifically invalidated all-zero numbers

BE0000000000 would previously result in a valid VAT number


arthurdejong:
Thanks for your fix. I've merged it as 3a592e4 with the addition of a test case.

Issue #125, opened 2019-06-09
Add support for Liechtenstein TIN


unho:
* https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/LIECHTENSTEIN%20TIN.pdf
* According to https://www.llv.li/inhalt/117378/amtsstellen/arbeitgeber-peid-re the "Handelsregister Nummer" (which has the form of **FL-0002.458.973-3**) can be converted to the PEID number by dropping the leading **FL-** as well as the leading zeroes and the last digit, like **2458973**.
  * This is also stated in https://www.liechtenstein-business.li/en/Service-for-entrepreneurs/After-founding-the-company/personnel/employment-registration

Online search:
* https://www.oera.li/cr-portal/suche/suche.xhtml

Examples:
* https://griffin.li/imprint/?lang=en
  > 1053541
* https://shop.timeless.li/Imprint
  > 2467526
* http://www.projekt365.li/
  > 2518516
* https://www.brainbot.li/
  > 2555984
* https://gross-steg.li/wp-content/uploads/2013/10/2012-Viehtrieb-Steg-V4b.pdf
  > 2766
84167
84166
* https://www.erminson.li/impressum.html
  > 2536979
* https://www.sieunder.li/impressum.html
  > 2580190
* http://www.jankowski.li/
  > 2557953
* https://www.hartmannrealestate.li/%C3%BCber-uns/
  > 2547594
* https://www.one-commerce.li/kontakt.htm
  > 2031108
* http://abgb.li/ECG.html
  > 2385259
* https://www.granville.li/impressum
  > 2.580.073
* https://future-marketing.li/impressum/
  > 2581203
* https://www.lanv.li/Portals/0/2018/Volksblatt%20gek%C3%BCndigte%20haben%20es%20schwer.pdf
  > 100364
* https://www.vaduz.li/fileadmin/vaduz-li/Politik-Verwaltung/Kurzprotokolle/2019/OEffentliches_Protokoll_22_01_2019.pdf
  > 2043832
* More PEID numbers can be obtained from the "Handelsregister Nummers" https://www.google.com/search?q=site%3A.li+%22Handelsregister+Nummer%22

PR #241, opened 2020-10-19
Li peid support

ramsondon:
Added support for Liechtenstein PEID (Personenidentifikationsnummer) validation

### DIFF
--- a/stdnum/li/__init__.py
+++ b/stdnum/li/__init__.py
@@ -1,0 +1,24 @@
+# __init__.py - collection of Liechtenstein numbers
+# coding: utf-8
+#
+# Copyright (C) 2020 Matthias Schmid
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA
+
+"""Collection of Liechtenstein numbers."""
+
+# provide aliases
+from stdnum.li import peid as vat  # noqa: F401
--- a/stdnum/li/peid.py
+++ b/stdnum/li/peid.py
@@ -1,0 +1,86 @@
+# peid.py - library for Liechtenstein Personenidentifikationsnummer
+#
+# Copyright (C) 2020 Matthias Schmid
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA
+
+"""PEID (Liechtenstein tax code for individuals and entities).
+
+The Personenidentifikationsnummer (PEID) is an numeric code up to 12
+digits used to identify entities and individuals residing in
+Liechtenstein.
+
+The first digits might be omitted if they correspond to the numeral
+zero. Nevertheless the minimum number of digits is four.
+
+
+More information:
+
+* http://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/
+* https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/LIECHTENSTEIN%20TIN.pdf
+
+>>> compact('00001234567')
+'1234567'
+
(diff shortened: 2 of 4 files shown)


stdnum/li/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA
+
+"""Collection of Liechtenstein numbers."""
+
+# provide aliases
```
unho, line 23: Please provide an alias to `vat`.

unho:
Can you please squash all these commits into a single one?

unho:
Please include a reference to issue #125

tests/test_li_peid.doctest
```diff
@@ -0,0 +1,68 @@
...
+These have been found online and should all be valid numbers.
+
+>>> numbers = '''
+...
+... 0001339050
+... 0843089848
+... 000012345678
+... 12345678
+... 001234567890
+... 123456789012
```
unho, line 79: Please also include the PEID numbers listed in issue #125

ramsondon: Thanks for your comments.. I'll do that. 
What about the Handelsregisternummer for employers. Shall we automatically compact this number "FL-0002.123.456-7" to a PEID?
or would you recommend to make a second implementation then?

unho: IMHO a separate implementation would be ideal.

stdnum/li/__init__.py
ramsondon: alias has been provided

tests/test_li_peid.doctest
ramsondon: all test cases from #125 have been included except the Handelsregisternummer. I would suggest to create a new Issue for this and a separate pull request.

unho:
Adjust commit message to indicate that it fixes issue #125.

stdnum/li/peid.py
```diff
@@ -0,0 +1,87 @@
...
+
+    return number
+
+
+def is_valid(number):
+    """Check if the given fiscal code is valid."""
+    try:
+        return bool(validate(number))
+    except ValidationError:
+        return False
```
unho, line 87: Add a `format` function as well.

stdnum/li/peid.py
```diff
@@ -0,0 +1,87 @@
...
+    return clean(number).strip().lstrip('0').replace('.', '', 2)
+
+
+def validate(number):
+    """Check if the given fiscal code is valid. This checks the length and
+    whether the check digit is correct."""
+    number = compact(number)
+    if len(number) < 4 or len(number) > 12:
+        raise InvalidLength()
+    if not _code_re.match(number):
```
unho, line 76: Use `isdigits` instead of a regex.

stdnum/li/peid.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+# regular expression for matching Liechtenstein Personenidentifikationsnummer
+_code_re = re.compile(r'^[0-9]{4,12}$')
+
+
+def compact(number):
+    """Convert the number to the minimal representation. This strips the
+    number of any valid separators and removes surrounding whitespace."""
+    return clean(number).strip().lstrip('0').replace('.', '', 2)
```
unho, line 67: You can pass `clean` the characters you want to get removed, like `clean(number, ' .')` which will remove both the whitespace and the period characters, so the `strip` and the `replace` are unnecessary. Note that the `lstrip` is still required.

unho: On a second thought maybe keeping `strip` is not a bad idea since it would remove newlines and such.

tests/test_li_peid.doctest
```diff
@@ -0,0 +1,103 @@
...
+... 2.580.073
+... 2581203
+... 100364
+... 2043832
+...
+... '''
+>>> [x for x in numbers.splitlines() if x and not peid.is_valid(x)]
+[]
+
+>>> peid.compact("00345345")
```
unho, line 102: Adjust this check to validate `format` (which is missing), and move it after line 62.

stdnum/li/peid.py
ramsondon: I have been using `.replace('.', '', 2)` instead of `clean(number, ' .')` because I wanted to make sure only 2 dots are being removed. as if 3 or more would be there it would be invalid according to the samples of #125 . as far as I know according to the spec of the PEID there is actually no "dot" in at all.

unho: I don't see any problem with removing all the dots.

tests/test_li_peid.doctest
ramsondon: implemented format with test case

stdnum/li/peid.py
ramsondon: replaced regex with isdigits

stdnum/li/peid.py
ramsondon: implemented format function as well

stdnum/li/peid.py
ramsondon: fixed compact according to @unho 's suggestions

arthurdejong:
Hi @ramsondon,

I've merged your PR as 5082af4. Thanks for providing the format, a petty that there was no check digit.

The PEID does not appear to be the Liechtenstein VAT number so I've not merged that part.

In Liechtenstein the VAT number appears to be the Mehrwertsteuer (MWST) which seems to be 5 digits long. There also seem to be some kind of registration number that seems to start with "FL", followed by 10 digits which some sources suggest is a PEID (although the PEID can be longer?) and a single digit (perhaps a check digit?).
There also seems to be a Legal Entity Identifier (LEI), Unternehmensnummer and Firmennummer but some of these could refer to the same number.

unho:
@arthurdejong it looks like you have not picked the `format` implementation. I am curious about the reason?

arthurdejong:
I generally only include a `format()` function if it changes anything to the number. For example if numbers have commonly used separators in certain places but they are more or less optional the `format()` function would return a number that includes the separators. If it doesn't add anything I leave it out because it is less code to maintain (albeit very little).

ramsondon:
Thx for merging... happy to have the peid now in the package!

PR #246, opened 2021-01-04
use abn as australian VAT number and remove GB from EU member states

viggo-devries:


arthurdejong:
In [PR 167](https://github.com/arthurdejong/python-stdnum/pull/167) @ahaeber indicated that the TFN was used as VAT number but [Wikipedia](https://en.wikipedia.org/wiki/VAT_identification_number#VAT_numbers_of_non-EU_countries) mentions the ABN. 
The [TFN Wikipedia page](https://en.wikipedia.org/wiki/Tax_file_number) also suggests that the ABN is assigned more often and can be used. Do you know of an official source that states that the ABN is equivalent to VAT numbers as used in the EU (GST in Australia)?

viggo-devries:
@arthurdejong I found these links: 

https://info.australia.gov.au/information-and-services/money-and-tax/tax/abn-australian-business-number
This is the ABN number and is used for VAT returns from the country etc.

https://www.ato.gov.au/individuals/tax-file-number/
This is the TFN number. It is a personal number not a business number.
"""You don't have to have a TFN, but without one you pay more tax. You also won't be able to apply for government benefits, lodge your tax return electronically"""

arthurdejong:
[The Australian Business Register](https://www.abr.gov.au/business-super-funds-charities/applying-other-registrations/apply-tfn-business) also mentions that the TFN can be assigned to businesses. But I did find [tax invoice requirements from the Australian Taxation Office](https://www.ato.gov.au/Business/GST/Tax-invoices/) that specifically mentions the ABN as being the number that you need to put in invoices. That makes it equivalent to a VAT number, fixed in cc3a970.

arthurdejong:
I'm looking into the UK VAT number changes that are also part of this PR. Can you provide some valid XI numbers that validate according to https://ec.europa.eu/taxation_customs/vies/ ? None of the numbers [in this PR](246/files#diff-822831e620fc369bfd3447a2031f3e1abfe5ca4a496ea47086d9c12881dc8b0dR698) appear to be valid when passing them to the VIES check.

PR #258, opened 2021-03-03
Fix typo

jwilk:


arthurdejong:
Thanks for the fix, merged as 7e69090.

PR #260, opened 2021-03-07
Update RNC the wsMovilDGII DOWN

jeffryjdelarosa:
the wsMovilDGII always is slow and is down, almost everytime the wsMovilDGII has issue.


jeffryjdelarosa:
It's Back Again

arthurdejong:
Perhaps we can add this alongside the SOAP-based lookup? Web scraping is really error prone and small changes on the site can make the code break.

PR #261, opened 2021-03-10
Add ISRC (International Standard Recording Code)

nuno-andre:
Hi!

This PR adds support for the [ISRC (International Standard Recording Code)](https://en.wikipedia.org/wiki/International_Standard_Recording_Code).

I've also added a little function to `util.py` that I often use to split strings into different sizes. I think it could be useful for other codes. If you think that it's misplaced or misnamed I have no problem in modifying the commit.
```
>>> chunks = 3, 4, 5, 4
>>> segment('0123456789ABCDEF', *chunks)
['012', '3456', '789AB', 'CDEF']
```

Also, comment that I'm writing a wrapper to use _python-stdnum_ with _Pydantic_ ([_Pydentic_](https://github.com/nuno-andre/pydentic)). I have a little library with some validators for Pydantic models, and I've found that yours has almost all of them (and many more), so I think it's way more productive to add the missing ones to _python-stdnum_ and wrap it up. Currently, it is little more than a proof of concept but I will develop it in the next few days, and will also send some more PRs if you like.

Thanks for your work, it's really amazing. I wish I had known about it sooner :)

### DIFF
--- a/README
+++ b/README
@@ -127,6 +127,7 @@ Currently this package supports the following formats:
  * ISIL (International Standard Identifier for Libraries)
  * ISIN (International Securities Identification Number)
  * ISMN (International Standard Music Number)
+ * ISRC (International Standard Recording Code)
  * ISO 11649 (Structured Creditor Reference)
  * ISO 6346 (International standard for container identification)
  * ISSN (International Standard Serial Number)
--- a/docs/stdnum.isrc.rst
+++ b/docs/stdnum.isrc.rst
@@ -1,0 +1,5 @@
+stdnum.isrc
+===========
+
+.. automodule:: stdnum.isrc
+   :members:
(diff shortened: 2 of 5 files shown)


arthurdejong:
Hi @nuno-andre,

Thanks for providing the number format. I've merged it as 36d723c. I took out the `segment()` function for now because it was slightly easier to switch to a regex than to turn it into something that also works for Python 2.7 (only recently support for Python 2.6 was dropped).

If there are any formats missing I would welcome pull requests but creating issues with a request for the format is also fine.

In general, python-stdnum is meant to be used in situations like with Pydantic (didn't know about that library before). I recommend calling `compact()` on a number before storing it in a database or similar storage because it tries to ensure a consistent for for the number so that it is easier to do queries for the number.

Thanks!

PR #265, opened 2021-05-26
Add CC (Número de Cartão de Cidadão, Portuguese Identity number)

davidmgvaz:
This adds CC (Número de Cartão de Cidadão, Portuguese Identity number)

### DIFF
--- a/stdnum/pt/cc.py
+++ b/stdnum/pt/cc.py
@@ -1,0 +1,106 @@
+# cc.py - functions for handling Portuguese Identity numbers
+# coding: utf-8
+#
+# Copyright (C) 2021 David Vaz
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA
+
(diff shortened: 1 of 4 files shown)


stdnum/pt/cc.py
```diff
@@ -0,0 +1,106 @@
...
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA
+
+"""CC (Número de Cartão de Cidadão, Portuguese Identity number).
+
+The Portuguese Identity Number is alfa numeric with the following format:
```
DimitriPapadopoulos, line 23: alphanumeric

stdnum/pt/cc.py
```diff
@@ -0,0 +1,106 @@
...
+        value_sum += value
+
+    if value_sum % 10 != 0:
+        raise InvalidChecksum()
+
+    return number
+
+
+def is_valid(number):
+    """Check if the number is a valid VAT number."""
```
DimitriPapadopoulos, line 96: ... a valid cartao de cidadao number

stdnum/pt/cc.py
davidmgvaz: Correct in last coomit

stdnum/pt/cc.py
davidmgvaz: Corrected in last commit

davidmgvaz:
Hi @DimitriPapadopoulos,

I have already corrected the two typos you found and also split the calck_check_digit to an autonomous function, as I saw that was the standard practice

arthurdejong:
Hi @davidmgvaz 

Thanks for the contribution. Could you provide some test numbers that can be used to confirm the implementation? For most number formats we have a number of tests with valid numbers to be able to validate the correctness of the calculations.

arthurdejong:
Also a part of the number is a Número de Identificação Civil which has its own check digit. Do you have any idea (or pointers) how to validate that number?

magamig:
@arthurdejong any particular reason to not update the README?

arthurdejong:
The README and documentation are updated as part of the release process with [update/numlist.py](https://github.com/arthurdejong/python-stdnum/blob/master/update/numlist.py).

PR #266, opened 2021-06-09
Improve NIF validation

DimitriPapadopoulos:


arthurdejong:
Hi @DimitriPapadopoulos ,

I merged your change as eeaf665 with a small change: the incorrect start digit will raise InvalidComponent instead of InvalidFormat (InvalidComponent is a little bit more descriptive).

PR #271, opened 2021-08-20
Add swedish postcode validator

cicuz:
As the title says, this PR adds validation for swedish postal codes


arthurdejong:
Hi @cicuz 

Thanks for your contribution. There are a few issues that should be resolved before merging:

- The copyright notice should match your name (in any case the copyright owner so that could also be your employer in some cases) and the current year
- The `validate()` function is expected to return a compact representation of the number (it currently returns a number with spaces in it, that should go to a separate `format()` function)
- Please either name the module `postal_code` (this matches the name of similar modules within python-stdnum) or something like postnummer which seems to be the name that it number has in Sweden

Thanks

cicuz:
Hello @arthurdejong,

I updated the code as you suggested; I was using the `nl` module as prototype, and that one is returning a formatted string from the `validate()` method, even though the situation there is a bit more complicated as the Dutch post codes contain letters as well.

By the way, do you think it's better to return the value as a number or as a string?

Thank you for your input and patience.

cicuz:
I will be adding tests too, getting accustomed to `doctest` which I didn't know before

cicuz:
Ok done, sorry for the spam 😅 

Let me know if it's looking alright,

cheers!

arthurdejong:
Hi @cicuz,

Thanks for the updated PR. I've merged it as 8071444.

I made a few small modifications. In general, postal codes should most likely not be stored as integers because usual mathematic operations on them do not make sense (e.g. adding two postal codes together or multiplying them). I also split the format validation from the length validation and simplified the format() function a bit.

Thanks!

cicuz:
That's awesome, thank you for the help, guidance, and patience :)

Do you think it'd be possible to package it into a new release?

PR #272, opened 2021-09-09
Typo found by codespell

DimitriPapadopoulos:
@arthurdejong  I cannot decide whether **Unknwon** has been propagated from an upstream source or is a typo. Can you help? See 40fcc24.


arthurdejong:
Hi @DimitriPapadopoulos 

Thanks for pointing this out. Most of the data files are imported from various sources (see the scripts in the update directory). This typo originated from https://en.wikipedia.org/wiki/Mobile_Network_Codes_in_ITU_region_3xx_(North_America) which I've fixed there.

The data files are updated on each release so the next release should have the updated files.

PR #277, opened 2021-09-28
Update existing modules for Indian stdnum + Add support for GSTIN and EPIC

vairag22:
- Updated existing modules for handling Aadhaar and PAN
- Added support for GSTIN (Indian VAT number)
- Added support for EPIC number (Indian Voter ID number)

### DIFF
--- a/stdnum/ec/ruc.py
+++ b/stdnum/ec/ruc.py
@@ -2,7 +2,7 @@
 # coding: utf-8
 #
 # Copyright (C) 2014 Jonathan Finlay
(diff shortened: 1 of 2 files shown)


vairag22:
pff.. finally done with tests.

vairag22:
Since I'm not very familiar with python tests I was unable to set up them properly, hence the ridiculous number of commits. Nevertheless I fixed all errors with the last commit, all checks passed. But this could've easily been a 4 commit PR, so should I do a new `git clone` and copy-paste latest changes or keep this as it is?

PR #284, opened 2021-11-28
Add Belgian National Number

cedk:
The Belgian National Number validation is described on https://fr.wikipedia.org/wiki/Num%C3%A9ro_de_registre_national


arthurdejong:
Sorry it took so long to review and merge this. It was merged as dcf4730.

PR #288, opened 2022-02-05
Compute birth date from Belgian National Number

cedk:
The Belgian National Number is most of the time build from the birth date (see https://fr.wikipedia.org/wiki/Num%C3%A9ro_de_registre_national#Composition).
So it may be useful to be able to extract the date from the number.


arthurdejong:
Hi @cedk thanks for your contribution. I've merged it as bda2a9c with some minor clarifications.

PR #294, opened 2022-04-08
gs1_128: add support for date parsing without day

alexis-via:
Date such as '(17)260400' is now properly interpreted as April 30th 2026.
Previous behavior: crash.

Information source : GS1 General specifications https://www.gs1.org/docs/barcodes/GS1_General_Specifications.pdf
section "Expiration date: AI (17)" page 162
```
When it is not necessary to specify the day (the day field is filled with two zeroes), the
resultant data string SHALL be interpreted as the last day of the noted month including any
adjustment for leap years (e.g., “130200” is “2013 February 28”, “160200” is “2016 February
29”, etc.).
```


pokoli:
It is needed that the encode uses such format or this is an optional value? 

I see your tests are failing because the dateutil library is not available. Could you please fix that?

alexis-via:
Tests are now green. I removed the dependency on dateutil.

arthurdejong:
@alexis-via thanks for providing this change and tanks @pokoli for providing some feedback!

Merged as 7d81eac.

Issue #295, opened 2022-04-13
Czech Bank Account Code and Number


petrprikryl:
Hi, would it be possible to add Czech bank account codes? 
https://www.cnb.cz/export/sites/cnb/cs/platebni-styk/.galleries/ucty_kody_bank/download/kody_bank_CR.pdf

And Czech bank account number? 
https://www.penize.cz/osobni-ucty/424173-tajemstvi-cisla-uctu-klicem-pro-banky-je-11 
http://www.zlatakoruna.info/zpravy/ucty/cislo-uctu-v-cr

Are you interested to integrate this if I will create PR? Thanks for the effort :wink:

petrprikryl:
Thanks for detailed guidance. PR is ready :wink:

PR #296, opened 2022-06-08
Added Czech bank account numbers

petrprikryl:
https://github.com/arthurdejong/python-stdnum/issues/295


arthurdejong:
Hi @petrprikryl,

Thanks for the PR. I've merged it as c5595c7 with a few small modifications. Some things could be simplified a bit and I've moved the zero padding to the `compact()` function. It strictly speaking doesn't make the number more compact but at least it ensures a consistent representation of the number (another approach would have been to trim leading zeros if possible).

PR #299, opened 2022-08-06
Small typo

vovavili:
Improper inflection of plurals in stdnum.isin.


arthurdejong:
Thanks for providing the fix!

PR #301, opened 2022-08-08
Upgrade GitHub Actions and Python versions

cclauss:
Python 3.11 release candidate 1 has just shipped TODAY at https://www.python.org/download/pre-releases so let's make sure the tests pass with it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,13 +12,13 @@ jobs:
   test:
     runs-on: ubuntu-latest
     strategy:
-      matrix:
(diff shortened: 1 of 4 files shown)


arthurdejong:
Hi @cclauss. I've merged two parts of this PR as b36c0d6 (the upgrades to the Github actions) and 351be74 (adding support for Python 3.10).

I'm a bit reluctant to officially support an as yet unreleased Python version. You can run the tests fine already (e.g. using `tox -e py311`). If there were things that required fixing in python-stdnum to support Python 3.11 or newer or if there were backwards incompatible or breaking changes it might be more important to add tests for a version that was not released yet.

If you feel any other parts of this PR should still be merged, feel free to open new PRs for them though.

PR #302, opened 2022-08-08
flake8: Either ignore long lines or set max-line-length but not both!

cclauss:
The remaining flake8 errors are fixed in #301


arthurdejong:
Thanks for pointing this out. I fixed this in 7ee0563 by specifying the long line ignores in the places where they were needed instead of globally.

PR #303, opened 2022-08-08
Fix typos discovered by codespell

cclauss:
https://pypi.org/project/codespell

The failing flake8 tests are fixed in #301


arthurdejong:
Hi @cclauss. Thanks for your contribution. However, the ChangeLog is generated on release from the Git commit history so this is not really possible to fix without re-writing the Git commit history. Also note that spelling and other mistakes in the NEWS file (which is included in the generated documentation) will be fixed.

Issue #300, opened 2022-08-08
Add Kenya TIN number


unho:
It is called PIN (Personal Identification Number) and is issued by the KRA (Kenya Revenue Authority).

* https://www.kra.go.ke/individual/individual-pin-registration/learn-about-pin/about-pin
* https://org-id.guide/list/KE-KRA
* https://www.cyber.co.ke/how-to-get-kra-pin-number-certificate-using-kra-portal/
  * According to this it is 11 digits long.
* https://stackoverflow.com/questions/51556967/how-to-preg-match-kra-pin-no
* https://docs.appruve.co/appruve-identity/kenya/verify-kenya-kra-pin
* https://api-docs.dojah.io/docs/lookup-kenya-tin
* https://www.facebook.com/KRACare/posts/personal-identification-number-pin-is-a-number-which-identifies-a-person-for-pur/1110866462257692/
* https://m.facebook.com/KRACare/photos/tip-of-the-day-httpsitaxkragoke/1270332812977722/

Online checker:
https://itax.kra.go.ke/KRA-Portal/pinChecker.htm

Validation code:
* https://gist.github.com/NimzyMaina/c32e6dc0ca9d6b83c6d28e3e25065677

Examples:
* A004416331M
* P051365947M
* P000594434U
* https://dennykinsassociates.com/kra-pin-numbers-for-all-companies-in-kenya/
  * P000619379O
  * P000609147B
  * P000595101M
  * P000594873H
  * P000608609Y
  * P000605255W
  * P000591260N
  * P000591376W
  * P051126254X
  * P000609718Z
  * P000594487J
  * P000599642I
  * P000615147U
  * P000619594P
  * P000608167F
  * P000609671I
  * P000597382L
  * P000598207Z
  * P000596967Q
  * P051099534I
  * P051116151S
  * P000597694S
  * P000597676Q
  * P000591115C
  * P000597819E
  * P051117306P
  * P000615058Y
  * P000608932A
  * P051124533N
  * P051123493Y
  * P051106014G
  * P051102575T
  * P051159129C
  * P051094112J
  * P000592089D
  * P000591575X
  * P000605619R
  * P051120393R
  * P000611519E
  * P051141902D
  * P000604053O
  * P000595728D
  * P051097904C
  * P000592118K
  * P051110833H
  * P000598418B
  * P051132776Z
  * P000592851X
  * P051127959L
  * P000602904I
  * P000598609F
  * P000603130E
  * P051092962C
[...]

PR #310, opened 2022-09-03
Add Kenyan TIN

unho:
Closes https://github.com/arthurdejong/python-stdnum/issues/300


arthurdejong:
Thanks for providing the format and doing all the research into this.

PR #318, opened 2022-09-09
Add more alias for stdnum.vatin to work for more countries

unho:


stdnum/li/__init__.py
```diff
@@ -19,3 +19,6 @@
 # 02110-1301 USA
 
 """Collection of Liechtenstein numbers."""
+
+# provide aliases
+from stdnum.li import peid as vat  # noqa: F401
```
arthurdejong, line 24: According to https://www.vatify.eu/liechtenstein-vat-number.html the MWST is issued for VAT purposes.

stdnum/md/__init__.py
```diff
@@ -19,3 +19,6 @@
 # 02110-1301 USA
 
 """Collection of Moldavian numbers."""
+
+# provide aliases
+from stdnum.md import idno as vat  # noqa: F401
```
arthurdejong, line 24: According to https://support.wix.com/en/article/about-vat-in-moldova the VAT number consists of 7 digits (with an optional MD prefix) which doesn't match the format of the IDNO.

stdnum/ua/__init__.py
```diff
@@ -19,3 +19,6 @@
 # 02110-1301 USA
 
 """Collection of Ukrainian numbers."""
+
+# provide aliases
+from stdnum.ua import edrpou as vat  # noqa: F401
```
arthurdejong, line 24: According to https://en.wikipedia.org/wiki/VAT_identification_number the Ukrainian is called ІНПП and consists of 12 digits which doesn't match the ЄДРПОУ implementation.

stdnum/za/__init__.py
```diff
@@ -19,3 +19,6 @@
 # 02110-1301 USA
 
 """Collection of South Africa numbers."""
+
+# provide aliases
+from stdnum.za import tin as vat  # noqa: F401
```
arthurdejong, line 24: According to https://support.upwork.com/hc/en-us/articles/15157294433683-South-Africa-VAT-for-Clients the VAT number is a 10 digit number starting with a 4. The existing TIN is also 10 digits but can't start with a 4.

arthurdejong (changes requested):
Please see the other comments

PR #319, opened 2022-09-09
Alter vat alias for India

unho:
According to https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/India-TIN.pdf the "Permanent Account Number" is the TIN number.

### DIFF
--- a/stdnum/in_/__init__.py
+++ b/stdnum/in_/__init__.py
@@ -21,4 +21,4 @@
 """Collection of Indian numbers."""
 
 # provide aliases
-from stdnum.in_ import gstin as vat  # noqa: F401
+from stdnum.in_ import pan as vat  # noqa: F401


arthurdejong:
According to https://en.wikipedia.org/wiki/VAT_identification_number#VAT_numbers_of_non-EU_countries the GSTIN is the VAT number and according to https://en.wikipedia.org/wiki/Goods_and_Services_Tax_(India) GST has replaced VAT in 2017.

https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/India-TIN.pdf does not talk about VAT but income tax.

While GST is not exactly VAT it appears to be closer than income tax.

Closing this but feel free to re-open if there is more information.

unho:
@arthurdejong I do realize that. But if you want to go that route, then a new `tin` (or `vatin`) alias should be added so `stdnum.vatin` works. Also this will make third party software, like Odoo, to not properly work since they use stdnum `vat` aliases to identify the TIN validation for the countries.

This is something we have already talked a lot about, for example in https://github.com/arthurdejong/python-stdnum/pull/324#issuecomment-1528980449 and https://github.com/arthurdejong/python-stdnum/pull/316

PR #324, opened 2022-09-10
Add vat alias so stdnum.vatin works for the United States

unho:
While stdnum provides several modules implementing some sort of TIN for the United States, EIN is the only only used for companies and institutions.

### DIFF
--- a/stdnum/us/__init__.py
+++ b/stdnum/us/__init__.py
@@ -19,3 +19,6 @@
 # 02110-1301 USA
 
 """Collection of United States numbers."""
+
+# provide aliases
+from stdnum.us import ein as vat  # noqa: F401


arthurdejong:
I'm pretty sure the US does not have the concept of VAT but they have sales tax instead: https://en.wikipedia.org/wiki/Value-added_tax#United_States

Closing this for now. We may want to introduce something along the lines of validation of any valid TIN but that can get complicated pretty fast and which kind of identifiers are appropriate is most likely very application-dependant.

unho:
@arthurdejong The validation of any valid TIN is provided by `stdnum.vatin` module, which relies on these `vat` aliases. So this PR was just adding this tiny change to make sure these module was able to validate TIN numbers for the United States.

If you prefer we can instead use a different alias, but then we will have to review all the countries to untagle the VAT number and TIN number, if that is even possible.

PR #338, opened 2022-10-16
Adding support for Slovenian EMŠO (Unique Master Citizen Number)

bblaz:


arthurdejong:
Hi @bblaz,

Thanks for your contribution. I've merged it as feccaff. I've compacted the module documentation a bit, made the check digit calculation simpler and made the code more follow the layout of other modules. I've removed the extract() function because we don't have that for the other modules.

PR #347, opened 2023-01-05
Extract gender from Belgian National Number

jeffh92:


arthurdejong:
Hi @jeffh92,

Thanks for your contribution. It was merged as 42d2792 (with only minor modifications).

PR #395, opened 2023-02-12
Add support for Senegal TIN

### LINKED ISSUE #357: Add support for Senegal TIN
* http://www.dci-sn.sn/index.php/obtenir-son-ninea
* It looks like it is called `Numéro d'Identification national des Entreprises et Associations (NINEA)`

* Online search: https://www.creationdentreprise.sn/rechercher-une-societe

Examples:
* `30672212G2` from https://www.camacoes.sn/index.php/es/listado-de-socios/308-produmel
* `008895677` from https://creationdentreprise.sn/en/node/38945
* Dozens on https://creationdentreprise.sn/en/finding-business (requires clicking on the details page for each company on the list)
* https://www.google.com/search?q=site%3Awww.camacoes.sn+%22NINEA%22

unho:
Fixes #357

### DIFF
--- a/stdnum/sn/__init__.py
+++ b/stdnum/sn/__init__.py
@@ -1,0 +1,24 @@
+# __init__.py - collection of Senegal numbers
+# coding: utf-8
+#
+# Copyright (C) 2023 Leandro Regueiro
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


arthurdejong:
Hi @unho,

Thanks for providing this and sorry for taking so long to merge it. It is merged with a few clean-ups as dd22123.

I also implemented check digit validation in d332ee1 It was quite easy once I saw the pattern in a few of the test numbers you provided. For example, there were two numbers that only differed in one digit but still valid. The difference between the differing digits was 5 which shouldn't happen if you do modulo 11 so I assumed it was modulo 10 and the weight for that digit was 2. Playing around a bit resulted in the alternating 1 and 2 weights.

There were only 3 numbers from the original test set that did not match the check digit algorithm and I confirmed on https://e-ninea.ansd.sn/search_annuaire that these were indeed invalid numbers.

PR #396, opened 2023-02-13
Add support for Finland HETU new century indicating signs

mjturt:
Finland's HETU updated on 2023-01-01 adding more century indicating sign options. Detailed information of the reformation can be found at the [Digital and Population Data Services Agency site](https://dvv.fi/en/reform-of-personal-identity-code).

The reformation doesn't affect checksum validation, and all the old HETU's are still valid.

### DIFF
--- a/stdnum/fi/hetu.py
+++ b/stdnum/fi/hetu.py
@@ -50,15 +50,15 @@ from stdnum.util import clean
 
 _century_codes = {
     '+': 1800,
-    '-': 1900,
-    'A': 2000,
 }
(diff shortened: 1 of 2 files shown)


arthurdejong:
Thanks for providing the fix. I've merged it as 36858cc.

Since all codes are still valid I don't think this is a breaking change (at least not for python-stdnum). What is important is that python-stdnum should not reject valid numbers.

mjturt:
Great, thank you! When do you plan to make a new release?

arthurdejong:
I hope to find the time to review a number of the open PR soon and make a release.

PR #402, opened 2023-03-01
Add 2020 decade Albania TIN

Dj0ulo:
The first letter of Albania TIN corresponds to the decade it was issued. (Source: https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/Albania-TIN.pdf)

This commit adds 'M' to the regex in order to match TIN from this decade.

### DIFF
--- a/stdnum/al/nipt.py
+++ b/stdnum/al/nipt.py
@@ -46,7 +46,7 @@ from stdnum.util import clean
 
 
 # regular expression for matching number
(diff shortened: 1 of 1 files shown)


arthurdejong:
Hi @Dj0ulo,

Thanks for pointing this out and providing the PR. I took this a little further and fixed it as a09a7ce. It turns out that the scope of the original implementation had some problems and the range we should accept is actually larger than the previous range implied.

PR #407, opened 2023-03-06
Cnp counties

RaduBorzea:


arthurdejong:
Hi @RaduBorzea,

After doing some research I found https://github.com/vimishor/cnp-spec/blob/master/spec.md which clarified things (though I'm not sure on how official that specification is).  In cf14a9f I merged your suggestion with some modifications (added tests, also made it part of the validation).

PR #409, opened 2023-03-09
Update database files

### LINKED ISSUE #399: IBAN Validation fails on Burundi, Djibouti, Russia, Somalia and Sudan
I'm using stdnum to validate IBAN in my own repo so to test it before going on production I make some tests with a list of IBANs from this [site](https://www.iban.com/structure): from this list I got 83 countries which 5 of them failed on testing. Is this expected? I tried other ways to validate these IBANs and are correct. My code is something like this:

```python
from stdnum.iban import is_valid
# For example in this case a failing number would be DJ2110002010010409943020008
def test(self, number):
  assert is_valid(number=number, check_country=True)

DimitriPapadopoulos:
Run [update/iban.py](https://github.com/arthurdejong/python-stdnum/blob/master/update/iban.py) to update [stdnum/isbn.dat](https://github.com/arthurdejong/python-stdnum/blob/master/stdnum/isbn.dat).

Update to _Release 93 – February 2023_ of the [IBAN Registry](https://www.swift.com/standards/data-standards/iban-international-bank-account-number), just adds Somalia.

Fixes #399.


arthurdejong:
Thanks, merged as bf1bdfe. I usually update the database files just before a release but if this helps I'll merge it now.

Issue #412, opened 2023-04-05
DeprecationWarning: pkg_resources is deprecated as an API


JiriKr:
It can be replaced with https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources .

Migration guide is available on https://importlib-resources.readthedocs.io/en/latest/migration.html .

arthurdejong:
Hi @JiriKr,

Thanks for the report and related fix. I don't see the deprecation warning when running the tests. Do you have an easy way to reproduce the warnings? Also, do you know why pkg_resources is deprecated?

pulse-mind:
Hi, Same deprecation on my side.

PR #413, opened 2023-04-05
Add importlib.resources as alternative for deprecated pkg_resources

JiriKr:
Issue #412


arthurdejong:
Hi @JiriKr,

I have solved the problem with a slightly different approach (7761e42) that also works with Python 3.7 and 3.8 (those versions have the `importlib.resources` module but don't have the `files()` function.

PR #418, opened 2023-06-20
Added Belgian BIS Number.

jeffh92:
This adds support for Belgian BIS numbers for non Belgians with social security rights.
It has the same format as Belgian national numbers, but can be differentiated by its month part, which is increased depending on whether the gender was known.

In my next PR, I will add a Belgian social security number (SSN) module that validates both national as BIS numbers, as the Begian SSN is the national number if available, otherwise a BIS number is assigned.
More information:
* https://www.socialsecurity.be/site_nl/employer/applics/belgianidpro/index.htm
* https://overheid.vlaanderen.be/personeel/regelgeving/insz-nummer

### DIFF
--- a/stdnum/be/bis.py
+++ b/stdnum/be/bis.py
@@ -1,0 +1,155 @@
+# coding=utf-8
+# bis.py - functions for handling Belgian BIS numbers
+#
+# Copyright (C) 2023 Jeff Horemans
+#
+# This library is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


arthurdejong:
Hi, thanks for providing this format. I've merged it as be33a80 with a few minor tweaks to the stdnum.be.nn module. Perhaps with a stdnum.be.ssn module we can simplify things further.

PR #424, opened 2023-12-08
Add European Excise Number

cedk:


cedk:
For now the check_seed tests fails because the EU service does not respect its WSDL.
I contacted them about that.

stdnum/eu/excise.py
```diff
@@ -0,0 +1,97 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA
+
+"""Excise Number
+
+The Excise Number is the identification number issued by the competent
+authority in respect of the person or premises.
```
arthurdejong, line 24: Can you add a bit more detail on what this number is and what it's used for?

stdnum/fr/accise.py
```diff
@@ -0,0 +1,95 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA
+
+"""n° d'accise.
```
arthurdejong, line 21: Can you improve this docstring a bit? For example: "n° d'accise (French number for ...)".

cedk:
Unfortunately the EU service is currently blocking the zeep library on this service.
Changing the `User-Agent`  fix it but there are also an error in the WSDL and the result of the service. The WSDL defines `subcategoryCode` as element when the service is using `subCategoryCode`.

I sent an email to EC-CENTRAL-HELPDESK@ec.europa.eu as their web page request.

PR #435, opened 2024-05-16
Add Partners Bank to banks.dat

DavidVrobel:
Add new czech bank


arthurdejong:
Thanks for reporting this. It was fixed in #439 with 91959bd.

Issue #107, opened 2019-06-02
Add support for Israel TIN


unho:
* I wasn't able to find the exact name (I found several with slight variations) so we will use the name referred in the government websites: מספר חברה (short version is ח.פ. )
* https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/Israel-TIN.pdf
* https://wiki.scn.sap.com/wiki/display/CRM/Israel (9 digits, check digit calculated using Luhn)
* https://he.wikipedia.org/wiki/%D7%AA%D7%90%D7%92%D7%99%D7%93
* https://www.avalara.com/vatlive/en/country-guides/africa-and-middle-east/israel/israel-vat-registration.html
  > Format | 123456789
9 characters
For incorporated businesses, the first digit is a 5
* According to https://en.wikipedia.org/wiki/VAT_identification_number:
  > 9 digit number. If the number of digits is less than 9, then zeros should be padded to the left side. The leftmost digit is 5 for corporations. Other leftmost digits are used for individuals. The rightmost digit is a check digit (using Luhn algorithm).

Algorithm:
* Algorithm to calculate Identity number (using Luhn) https://he.wikipedia.org/wiki/%D7%A1%D7%A4%D7%A8%D7%AA_%D7%91%D7%99%D7%A7%D7%95%D7%A8%D7%AA#%D7%A1%D7%A4%D7%A8%D7%AA_%D7%91%D7%99%D7%A7%D7%95%D7%A8%D7%AA_%D7%91%D7%9E%D7%A1%D7%A4%D7%A8_%D7%94%D7%96%D7%94%D7%95%D7%AA_%D7%91%D7%99%D7%A9%D7%A8%D7%90%D7%9C which according to https://he.wikipedia.org/wiki/%D7%AA%D7%90%D7%92%D7%99%D7%93 is the same algorithm used for TIN number.

Online search:
* https://www.gov.il/en/service/company_extract (links to page listed below)
* https://ica.justice.gov.il/GenericCorporarionInfo/SearchCorporation?unit=8 (this is only for companies, in other words it requires company number first two digits to be 51 through 59)
  * There are other online searchs for associations, etc.

Examples:
[...]

PR #436, opened 2024-05-17
Support both corporate and individual Israeli TINs for generic validation

odony:
According to the official references mentioned in #107, Israeli companies can use two types of Tax Identification Numbers / VAT codes:
- Corporate Registration Numbers for corporations, as currently implemented by `stdnum.il.hp`
- ID Card numbers for sole proprietorship businesses (individuals), as currently implemented by `stdnum.il.idnr`

For example, the [TIN rules that were provided to OECD by the Israeli tax administration](https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/Israel-TIN.pdf) state:
> The "TIN" for an Individual is his ID number.

Based on these references, it seems that the validation of a generic VAT number for Israel via `stdnum.il.vat` should allow both formats.

Further, considering that `stdnum.il.idnr` is a simple luhn validation that will also work for the numbers that `stdnum.il.hp` currently verifies, this can be done by simply changin the default alias to point to it instead. The only difference between the two algorithms is that the `stdnum.il.hp` one verifies that the first digit is 5.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,6 +35,8 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+        env:
+          PIP_TRUSTED_HOST: "pypi.python.org pypi.org files.pythonhosted.org"
       - name: Install dependencies
         run: python -m pip install --upgrade pip tox
       - name: Run tox


arthurdejong:
Hi, sorry it took a bit to respond. It is probably better to have one module that validates the VAT number (or TIN) separately. If both ID number and company number are valid TIN, is there a name for the TIN? For example, if either number is filled in in a form, what is it referred to?

PR #439, opened 2024-05-21
Update Czech database files

petrprikryl:
Updated `/stdnum/cz/banks.dat` using update script `/update/cz_banks.py`.

IMO better than #435.


arthurdejong:
Thanks, merged as 91959bd

Btw, the data files are re-generated as part of the release process so this is mostly for the benefit of those using the Git version.

Issue #444, opened 2024-07-04
Incorrect usage of timeout for zeep


jmak-odoo:
The current way that the timeout is set for zeep does not function properly.
The parameter for POST/GET timeouts is operation_timeout as discussed here: https://github.com/mvantellingen/python-zeep/issues/140
https://github.com/arthurdejong/python-stdnum/blob/e951daca447f678f5a3f3c70dcc535eb8d7449b4/stdnum/util.py#L256

PR #445, opened 2024-07-04
fix zeep client correct timeout

jmak-odoo:
The timeout parameter of the zeep transport class is not responsable for POST/GET timeouts. The operational_timeout parameter should be used for that.
See https://github.com/mvantellingen/python-zeep/issues/140

Closes https://github.com/arthurdejong/python-stdnum/issues/444


arthurdejong:
Hi @jmak-odoo, thanks for providing the fix. It is merged as 6cbb9bc.

PR #449, opened 2024-07-25
Added Dutch identiteitskaartnummer.

jeffh92:
Identiteitskaartnummer, Paspoortnummer (the Dutch passport number).

Each Dutch passport has an unique number of 9 alphanumerical characters.
The first 2 characters are always letters and the last character is a number.
The 6 characters inbetween can be either or.

The letter "O" is never used to prevent it from being confused with the number "0".
Zeros are thus allowed, but are not used anymore in numbers issued after December 2019.

More information:

* https://www.rvig.nl/node/356
* https://nl.wikipedia.org/wiki/Paspoort#Nederlands_paspoort

### DIFF
--- a/stdnum/nl/identiteitskaartnummer.py
+++ b/stdnum/nl/identiteitskaartnummer.py
@@ -1,0 +1,92 @@
+# identiteitskaartnummer.py - functions for handling Dutch Passport Number.
+# coding: utf-8
+#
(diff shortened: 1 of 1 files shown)


arthurdejong:
Hi @jeffh92,

Thanks for providing this module. I've merged it as 56036d and it will be in the next release.

PR #465, opened 2025-02-27
Implement the new ro.onrc identifier

### LINKED ISSUE #464: Support new Romanian ro_onrc format along with the old format
Example old format: J52/750/2012, new format is J2012000750528.
@dotbit1 to do: find the specification for the new format and how under what conditions the old format is valid.

dotbit1:
Closes #464;
Tested with multiple new format codes.

Used https://targetare.ro/blog/schimbari-importante-la-registrul-comertului-ce-trebuie-sa-stii-despre-noul-format-al-numarului-de-ordine as a source of info.

The part about using the J, F, C ASCII values is wrong, it should be the number of the letter, etc J = 10, etc.

### DIFF
--- a/stdnum/ro/onrc.py
+++ b/stdnum/ro/onrc.py
@@ -1,7 +1,7 @@
 # onrc.py - functions for handling Romanian ONRC numbers
 # coding: utf-8
 #
-# Copyright (C) 2020 Dimitrios Josef Moustos
+# Copyright (C) 2024 Dimitrios Josef Moustos
 # Copyright (C) 2020 Arthur de Jong
 #
 # This library is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


arthurdejong:
Hi @dotbit1,

Could you include the numbers you tested with?

dotbit1:
Hi Arthur,
The code I wrote makes the existing tests fail.
The way I find 100% valid codes is: go to https://mfinante.gov.ro/ro/web/site/info-pj-selectie-nume-si-judet,
write some text that is likely to have many company names, ex. 4 letters and search,
then check each out and see if they have the new format.

Below a few examples:
J2012000750528
J2007007766403
J1992002621040
J1994000536225
J1993001036400

PR #467, opened 2025-03-13
Add type hints

### LINKED ISSUE #433: Provide type hints for type checkers and language servers
While there are a lot of modules in this package, since the common interface is very simple from a type system perspective I think it would introduce little to no additional maintenance burden to include type hints and the majority of the initial hints could be added automatically for the common interface using something like `sed` or `awk`.

If there's any interest in maintaining inline type hints I'd be happy to create a pull request, if you'd rather not, I can go the typeshed route instead. I've already contributed type hints for various third party packages including Chameleon, WebOb, WTForms, uWSGI and gevent, just to name a few.

Daverball:
Closes #433

This adds inline type hints to the project. While trying to keep runtime impact to the absolute minimum, while still allowing for runtime introspection of type hints.

As a side effect this puts a hard limit of Python 3.7 as our minimum version, since older versions than 3.5 did not have support for any inline annotations at all and older versions than 3.7 did not have support for `from __future__ import annotations`. The experience of maintaining type hints is much worse without the latter, so it seems like a reasonable trade-off to me, given that 3.6 has been EOL since 2022.

Otherwise it would make more sense to ship third party stubs.

Type hints are only guaranteed to be correct for Python 3.9+, which also seems like a reasonable trade-off, since current versions of type checkers don't support older versions than that anyways.

If you want to be able to introspect type hints, `typing_extensions` needs to be installed, which is once again reasonable, given that every major runtime typing library depends on `typing_extensions` and static type checkers have access to it via typeshed. Also for runtime introspection, the minimum Python version is 3.10, due to the use of `|` for type unions, which also seems reasonable, given that people that make runtime use of annotations tend to use newer version of Python and that 3.9 will be EOL in October. But if this is unacceptable, we can replace the few places we use the operator with `typing.Union` and `typing.Optional`.

This also fixes a couple of minor bugs that were discovered thanks to the static analysis provided by mypy.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,6 @@ setup(
         'Operating System :: OS Independent',
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
@@ -77,8 +76,11 @@ setup(
         'Topic :: Text Processing :: General',
     ],
     packages=find_packages(),
-    install_requires=[],
-    package_data={'': ['*.dat', '*.crt']},
+    python_requires='>=3.7',
+    install_requires=[
+        'importlib_resources >= 1.3  ; python_version < "3.9"',
+    ],
+    package_data={'': ['*.dat', '*.crt', 'py.typed']},
     extras_require={
         # The SOAP feature is only required for a number of online tests
         # of numbers such as the EU VAT VIES lookup, the Dominican Republic
--- a/stdnum/_typing.py
+++ b/stdnum/_typing.py
@@ -1,0 +1,119 @@
+# _typing.py - module for typing shims with reduced runtime overhead
+#
+# Copyright (C) 2025 David Salvisberg
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA
+
+"""Compatibility shims for the Python typing module.
+
+This module is designed in a way, such, that runtime use of
+annotations is possible starting with Python 3.9, but it still
+supports Python 3.6 - 3.8 if the package is used normally without
+introspecting the annotations of the API.
+
+You should never import *from* this module, you should always import
+the entire module and then access the members via attribute access.
+
+I.e. use the module like this:
+```python
+from stdnum import _typing as t
+
+foo: t.Any = ...
+```
+
+Instead of like this:
+```python
+from stdnum._typing import Any
+
+foo: Any = ...
+```
+
+The exception to that rule are `TYPE_CHECKING` `cast` and `deprecated`
+which can be used at runtime.
+"""
+
+from __future__ import annotations
+
+
+TYPE_CHECKING = False
+if TYPE_CHECKING:
+    from collections.abc import Generator as Generator
+    from collections.abc import Iterable as Iterable
+    from collections.abc import Mapping as Mapping
+    from collections.abc import Sequence as Sequence
+    from typing import Any as Any
+    from typing import IO as IO
+    from typing import Literal as Literal
+    from typing import cast as cast
+    from typing_extensions import TypeAlias as TypeAlias
+    from typing_extensions import deprecated as deprecated
+
+    from stdnum._types import GSTINInfo as GSTINInfo
+    from stdnum._types import IMSIInfo as IMSIInfo
+    from stdnum._types import NumberValidationModule as NumberValidationModule
+    from stdnum._types import PANInfo as PANInfo
+else:
+    def cast(typ, val):
+        """Cast a value to a type."""
+        return val
+
+    class deprecated:  # noqa: N801
+        """Simplified backport of `warnings.deprecated`.
+
+        This backport doesn't handle classes or async functions.
+        """
+
+        def __init__(self, message, category=DeprecationWarning, stacklevel=1):  # noqa: D107
+            self.message = message
+            self.category = category
+            self.stacklevel = stacklevel
+
+        def __call__(self, func):  # noqa: D102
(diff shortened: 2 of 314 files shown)


setup.py
```diff
@@ -77,8 +76,11 @@
         'Topic :: Text Processing :: General',
     ],
     packages=find_packages(),
-    install_requires=[],
-    package_data={'': ['*.dat', '*.crt']},
+    python_requires='>=3.7',
+    install_requires=[
+        'importlib_resources >= 1.3  ; python_version < "3.9"',
```
Daverball, line 81: Type checkers don't have great support for the `try`/`except` style of providing backwards compatibility. So instead of relying on `pkg_resources`, which may be missing even for Python versions 3.7 and 3.8 if a tool like `uv` was used to setup the environment, it seemed better to just rely on the `importlib_resources` backport with the minimum version that supports the `files` function. That way we don't have a soft-dependency on `setuptools`.

stdnum/de/stnr.py
```diff
@@ -90,28 +92,28 @@ def _clean_region(region):
...
         return self._re.match(number)
 
-    def replace(self, f, b, u, p):
+    def replace(self, f: str, b: str, u: str, p: str) -> str:
         items = iter([f, b, u, p])
         return re.sub(r'([FBUP])\1*', lambda x: next(items), self._fmt)
 
 
 # Convert the structure to something that we can easily use
 _number_formats_per_region = dict(
```
arthurdejong, line 110: Is there a nicer solution for this other than leaving both `_number_formats_per_region_raw` and `_number_formats_per_region` around?

stdnum/_typing.py
```diff
@@ -0,0 +1,119 @@
...
+The exception to that rule are `TYPE_CHECKING` `cast` and `deprecated`
+which can be used at runtime.
+"""
+
+from __future__ import annotations
+
+
+TYPE_CHECKING = False
+if TYPE_CHECKING:
+    from collections.abc import Generator as Generator
```
arthurdejong, line 53: I don't really understand why this code is here. I don't think there is any situation where you can have this if branch be executed.

This code appears to have another origin and more generic than is needed for python-stdnum, where did it come from?

Daverball: This branch is for type checkers only, conversely the `else` branch is never looked at by type checkers.

The entire point of this file is to defer the `typing` import, since it is quite expensive, which would make `python-stdnum` less attractive in embedded environments. But the magic for this is too dynamic for type checkers to be able to understand it, so this branch basically summarizes for type checkers what `typing` symbols this module re-exports.

We could avoid the branch by shipping a separate `_typing.pyi` file, but it seems easier to keep the typing-only branch in sync with the runtime branch if it's all in the same file.

arthurdejong: I'm still struggling with getting this merged. The whole `_typing.py` file feels like a pretty big hack just to get some type annotations. I will probably merge some bits of this earlier. I'm in the process of extracting some parts into separate commits which leaves the huge one with just mechanical changes.

Are imports from `collections.abc` also expensive? If there is an efficient way to be able to reference `Any` we can probably make something that is has minimal runtime impact.

Daverball: `collections.abc` is not as bad as far as overhead is concerned I believe. I just think it's easier if all the types are collected in a common utility module when `python-stdnum` relies on such a small number of them. There's less confusion overall if there's only one place to import types from.

The only real friction I see with this approach is, that if you need a type that hasn't been added to `_typing.py` yet, you need to add it to both branches in the correct way. But this should be rare, given how much code there is and how little need there is for anything fancy typing-wise.

There are however alternative approaches. We could simplify things, if you don't care to support runtime introspection of type hints at all we could just put all the typing only imports in a `if TYPE_CHECKING` block at the start of each file that needs it. This should only really negatively impact users of something like beartype.

There are however still things with runtime effects like `cast` and `@deprecated`, so for those alone we do need something like `_typing.py`, so we can provide our own implementation that doesn't depend on `typing`, but for type checkers to be able to understand what these functions do, we need to pretend it's their implementation, not ours.

It is unfortunately what you have to do currently if you want to minimize the runtime impact of type annotations. If you don't care to minimize the runtime impact we can add `typing_extensions` as a dependency and import everything at runtime.

Other than that the only option with no runtime overhead is separate stub files. But then you lose both runtime introspection of the type hints and the ability to type check the implementation. So it will not help you find any bugs in `stdnum`, just in code that uses it. So at that point it would probably be more ergonomic to contribute the stubs to typeshed, than maintain them in this project and deal with the hassle of updating two separate files each time.

Daverball: Also just in case this wasn't clear: The `if TYPE_CHECKING:` part isn't the hack. `typing.TYPE_CHECKING` exists as a constant and it's very commonly used to get around limitations of the type system. Doing `TYPE_CHECKING = False` is the currently supported way to avoid the `typing` import, since type checkers will still treat it like `typing.TYPE_CHECKING` (`False` at runtime, `True` at type checking time).

The only real hack is the module-level `__getattr__` so we can avoid runtime imports in annotations, since we deferred their execution via `from __future__ import annotations`. So with this example:

```
import stdnum._typing as t

x: t.Any
```

What happens is that, `__annotations__` becomes `{'x': 't.Any'}`, so the code `t.Any` doesn't get executed until you (or a different dependency) runs `typing.get_type_hints` or `inspect.get_annotations` on your modules/functions which both use `eval` internally to execute the stringized type expressions, so that's the first time `stdnum._typing.__getattr__` gets called and the import happens.

A module-level `__getattr__` to avoid the `typing` import is admittedly less common, but there are some libraries that use that trick. What is a lot more common is the use of `TYPE_CHECKING = False` and not supporting runtime introspection of type hints.

The more commonly used style would look like this:

```
TYPE_CHECKING = False
if TYPE_CHECKING:
    from typing import Any

x: Any
```

In this case `typing.get_type_hints` and `inspect.get_annotations` would raise a `NameError`, since `Any` is never actually imported. Although PEP 649 will improve the situation for runtime introspection and add a mode where unresolvable names are replaced with `typing.ForwardRef`, so you get back `{'x': typing.ForwardRef('Any')}` instead of a `NameError` for the above.

Daverball:
@arthurdejong Take all the time you need. I merged the current state of master and resolved the merge conflicts, so the diff should now reflect the current state of affairs.

arthurdejong:
Hi @Daverball,

I've merged most of your proposed patch as 0e857fb. I left out the new types and typing modules and made the changes a little more minimal. I've also left out the more specialised return types for gender, and the functions that return dictionaries for now just to have something in place to start with. I feel it is best to start with a minimalistic approach, see how that works and iterate from there.

Thank you for your contribution and your help and patience in getting this merged. Also thanks for all the context and explanation, it will help me greatly in adding type checking support in other Python code.

Daverball:
@arthurdejong Looks great. Thanks for spending the time on transforming the changes to suit your preferences.

As a side note: I'd recommend adding `if TYPE_CHECKING:` to `coverage.report.exclude_lines`. That way you don't have to manually add a `pragma: no cover` to every type checking block, beyond that `assert_never[(]`, `raise NotImplementedError` and `if __name__ == .__main__.` are other lines i usually add to my config.

PR #470, opened 2025-05-05
Deleteding superfluous modulus operation

lucasicr:
Since ``` -(x % 11) % 11 ``` and ``` -x % 11 ``` give always the same result, the first modulus operation can be safely deleted.

In the same way in the following code, the modulus in the fist line can be deleted with no effect on the output:
```  python
    s = sum(w * int(n) for w, n in zip(weights, number)) % 11
    return str(-s % 11 % 10)
```


arthurdejong:
Thanks. Merged as 8b78f78.

PR #475, opened 2025-05-14
Define in setup.py the minimal Python version

### LINKED ISSUE #474: Install python-stdnum with Python 3.6 fails
Since 0e857fb python-stdnum requires at minimum Python 3.7 so the setup.py should reflect that with a proper `python_requires` otherwise pip tries to install the wrong version.

cedk:
Closes #474


cedk:
It will be great to yank the current release 2.0 otherwise old Python version will keep trying to install it.

PR #476, opened 2025-06-05
Add personal id and business id alias for Estonia

TheMohni:


arthurdejong:
Thanks. Merged as deb0db1.

PR #479, opened 2025-08-05
Fix submodule visibility in country modules

philmas:
Add explicit imports to __init__.py files for all country modules to make submodules visible via dir() function. This resolves the issue where submodules like stdnum.gb.nhs, stdnum.us.ssn, etc. were not discoverable through introspection despite being importable directly.

Changes:
- Added explicit 'from stdnum.{country} import {module}' statements
- Added '# noqa: F401' comments to suppress unused import warnings
- Ensures all submodules are now visible in dir() output
- Maintains backward compatibility for direct imports

Fixes module discoverability for all country-specific validation modules.

### DIFF
--- a/stdnum/ar/__init__.py
+++ b/stdnum/ar/__init__.py
@@ -25,3 +25,4 @@ from __future__ import annotations
 # provide aliases
 from stdnum.ar import cuit as vat  # noqa: F401
 from stdnum.ar import dni as personalid  # noqa: F401
+from stdnum.ar import cbu  # noqa: F401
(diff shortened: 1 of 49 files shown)


arthurdejong:
Hi @philmas,

Thanks for your suggestion. I don't think I fully understand the use case behind the change. Introspection works fine with `pkgutil.walk_packages()` and there is also already a `stdnum.util.get_number_modules()` function that can help you with that.

For example, to get all the UK number format modules, you can do:

```
>>> from stdnum.util import get_number_modules
>>> [module.__name__ for module in get_number_modules('stdnum.gb')]
['stdnum.gb.nhs', 'stdnum.gb.sedol', 'stdnum.gb.upn', 'stdnum.gb.utr', 'stdnum.gb.vat']
```
(by default it will list all the number formats)

PR #490, opened 2025-10-14
Add support for nace revision 2.1

kat-star-me:
The EU has recently released the new nace classification version 2.1. I have added a new module to support 2.1 (rather than overwriting the existing one for revision 2, as this is probably still widely used). 
The data was collected here: https://showvoc.op.europa.eu/#/datasets/ESTAT_Statistical_Classification_of_Economic_Activities_in_the_European_Community_Rev._2.1._%28NACE_2.1%29/downloads
I am attaching the script I used for reformatting this into the numdb format.

[prepare_nace21_dat.py](https://github.com/user-attachments/files/22901196/prepare_nace21_dat.py)

The file provided contains the label in all European languages, a future improvement could be to support for mutliple languages for the nace descriptions.


arthurdejong:
Hi @kat-star-me,

Sorry it took a bit but going back and forth I've decided to keep one Python module for NACE and add an extra argument to the functions to select a revision. I did change the default to use revision 2.1.

Thanks for your contribution. It was merged as 2cf475c.

kat-star-me:
Great, thanks!

PR #493, opened 2026-05-07
Update ES CAE with EHA/3482/2007 activity keys and BOE reference

quiqueporta:
Update Spanish CAE module to align with Orden EHA/3482/2007: refresh the BOE link, add the F7 and GP activity keys, fix some descriptions and the Bizkaia spelling, plus doctest coverage for the new keys.

### DIFF
--- a/stdnum/es/cae.py
+++ b/stdnum/es/cae.py
@@ -31,7 +31,7 @@ place.
 
 More information:
 
-* https://www.boe.es/boe/dias/2006/12/28/pdfs/A46098-46100.pdf
+* https://www.boe.es/boe/dias/2007/12/01/pdfs/A49509-49640.pdf
 * https://www2.agenciatributaria.gob.es/L/inwinvoc/es.aeat.dit.adu.adce.cae.cw.AccW?fAccion=consulta
 
 >>> validate('ES00008V1488Q')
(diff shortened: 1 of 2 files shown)


quiqueporta:
Note on the failing mypy job

The Test / tox_job (mypy) failure is not caused by this PR. I reproduced it locally and the exact same 11 errors are present on master.

The root cause is the recent release of mypy 2.0.0, which tox.ini pulls unpinned. Two breaking changes affect this repo:                                                                                                                                     
                                                                                                                                                                                                                                                               
1. python_version = 3.9 in setup.cfg is no longer accepted (mypy 2.0 requires 3.10+), which also makes the first tox command (mypy -p stdnum --python-version 3.9) fail outright.
2. Stricter var-annotated / no-any-return checks now flag the dynamic module-loading patterns in stdnum/iban.py, stdnum/vatin.py, stdnum/numdb.py, stdnum/util.py, stdnum/eu/vat.py and stdnum/eu/excise.py (e.g. country_modules, _open_databases,  _soap_clients).   

Possible fixes (out of scope for this PR, happy to send a separate one if useful):
- Pin mypy<2.0 in tox.ini as a short-term fix.
- Bump the declared python_version to 3.10 in setup.cfg and add explicit annotations to the dynamic-loading dicts.

This PR only touches stdnum/es/cae.py and its doctest, so it doesn't introduce any of these errors.

PR #495, opened 2026-05-17
Add RU SNILS

talveg-1:
Added an implementation for Russian SNILS numbers

### DIFF
--- a/stdnum/ru/snils.py
+++ b/stdnum/ru/snils.py
@@ -1,0 +1,75 @@
+"""СНИЛС (Страховой номер индивидуального лицевого счёта, Russian Individual insurance account number)
+
+More information:
+
+* https://en.wikipedia.org/wiki/SNILS_(Russia)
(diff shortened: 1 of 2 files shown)


arthurdejong:
Hi @talveg-1,

Thanks for providing this. I have some comments though:

- Copyright and license information is missing from the files
- Some tests are failing due to formatting problems
- It would be nice if there were real actual numbers in the tests. Do you know if there is a way to get some 10-20 real-world numbers?

I see the check digit algorithm can be found on
https://ru.wikipedia.org/wiki/Контрольное_число#Страховой_номер_индивидуального_лицевого_счёта_(Россия)

There does appear to be one additional check in that the check digit is only expected to be valid for numbers greater than 001-001-998 (if the translation I looked at is correct) implying that numbers lower than that don’t necessarily have to have a valid check digit.

The other issues are trivial to fix but if you can confirm the author name and that you agree to distribute this under the LGPLv2+ I can reasonably easily fix the other bits.

PR #498, opened 2026-06-16
Support GS1-128 date AIs with a four-digit year (7250, 7251)

binggao1230:
GS1 Application Identifiers 7250 (DOB) and 7251 (DOB TIME) carry a four-digit
year (`N8` = `YYYYMMDD`, `N12` = `YYYYMMDDhhmm`), but `gs1_128` only handled
two-digit-year date formats, so both AIs were broken in either direction:

- `info('725019800715')` raised a bare `ValueError: time data '19800715' does
  not match format '%y%m%d%H'`. The `N8` value fell through to the generic
  `'%y%m%d%H%M%S'[:len(value)]` branch, which parsed the four-digit year as `YY`.
- `info('7251198007151430')` raised `ValueError` because the 12-digit `N12`
  value was treated as two `YYMMDD` dates and `'198007'` is not a valid date.
- `encode({'7250': datetime.date(1980, 7, 15)})` raised
  `ValueError: unsupported format: N8`.

`_encode_date()` now formats `N8` as `%Y%m%d` and `N12` as `%Y%m%d%H%M`, and
`_decode_date()` parses them back with the matching four-digit-year formats.
`N12` is also removed from the two-date branch: the only date AI that encodes a
pair of `YYMMDD` dates is 7007 (`N6[+N6]`), whereas 7251 is a single datetime.

Round-trip doctests for both AIs are added to `tests/test_gs1_128.doctest`.

### DIFF
--- a/stdnum/gs1_128.py
+++ b/stdnum/gs1_128.py
@@ -122,8 +122,15 @@ def _encode_date(fmt: str, value: object) -> str:
         # Format date in different formats
         if fmt in ('N6', 'N6..12', 'N6[+N6]'):
             return value.strftime('%y%m%d')
+        elif fmt == 'N8':
+            # Date with a four-digit year (YYYYMMDD), e.g. AI 7250 (DOB).
+            return value.strftime('%Y%m%d')
         elif fmt == 'N10':
             return value.strftime('%y%m%d%H%M')
+        elif fmt == 'N12':
+            # Date and time with a four-digit year (YYYYMMDDhhmm), e.g. AI 7251
+            # (DOB TIME).
+            return value.strftime('%Y%m%d%H%M')
         elif fmt in ('N6+N..4', 'N6[+N..4]', 'N6[+N4]'):
             value = value.strftime('%y%m%d%H%M')
             if value.endswith('00'):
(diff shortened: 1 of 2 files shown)


binggao1230:
Updated the branch to pin `mypy<2` in the tox mypy environment. The failing CI was installing mypy 2.1, which no longer supports the existing `--python-version 3.9` target in this project. This keeps the current 3.9 type-check target working without changing runtime code.

Local verification:
- `uvx --python 3.13 tox -e mypy --skip-missing-interpreters false`
- `uvx --python 3.13 tox -e flake8 --skip-missing-interpreters false`
- `uvx --python 3.13 --with pytest --with pytest-cov pytest --doctest-modules --doctest-glob='*.doctest' stdnum/gs1_128.py tests/test_gs1_128.doctest`

binggao1230:
Rebased on master; I dropped my tox.ini mypy pin from this branch since you already capped mypy the same way in b99cfad, so the PR is now just the GS1-128 change.

arthurdejong:
Hi @gaoflow,

Thanks for providing this fix. I've merged it as 8a212c7.

PR #499, opened 2026-06-24
ro/cf: strip RO country prefix in compact() for consistent output

### LINKED ISSUE #485: Inconsistent behavior with country prefixes
As far as I can tell from a bit of research (although I'm not 100% sure), the domestic form of both the Polish NIP and Romanian CF numbers are digits only, and a country prefix can be optionally added to the beginning when using it as an EU VAT number.

The NIP validator behaves like I would expect it to - the documentation says it should return "a compact, consistent representation of the number or code", and it does so - it always returns the domestic form of the number, without the country prefix. The CF validator, however, returns whatever it was passed - if you pass in a country-prefixed number, it outputs a country-prefixed number, and vice versa for a domestic number.

It's hard to rely on the output of the library when for some countries the format of the output depends on the format of the input. I don't know if this is just an issue of the Romanian validator, or a wider issue, but I think it warrants being checked out and corrected, so that the output is indeed compact and consistent.

Small side note: ro/cf.py is also missing checksum validation

binggao1230:
## Description

Fixes #485. `stdnum.ro.cf` produced inconsistent output depending on whether the optional `RO` country prefix was present in the input:

```python
>>> validate('RO 185 472 90')
'RO18547290'          # before
>>> validate('185 472 90')
'18547290'
```

Both inputs denote the same number, but `compact()` did not strip the `RO` prefix the way the `cui.compact()` it delegates to (and other country modules such as `pl/nip`) do.

## Fix

- `compact()` now strips the `RO` prefix, returning the domestic form — matching `cui.compact()` and sibling modules.
- `validate()` is simplified: the manual `RO` split is removed since `compact()` now handles it.
- The docstring example is updated to reflect `'18547290'`.

`validate('RO 185 472 90')` and `validate('185 472 90')` now both return `'18547290'`.

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/stdnum/ro/cf.py
+++ b/stdnum/ro/cf.py
@@ -20,8 +20,8 @@
 
 The Romanian CF is used for VAT purposes and can be from 2 to 10 digits long.
 
->>> validate('RO 185 472 90')  # VAT CUI/CIF
-'RO18547290'
+>>> validate('RO 185 472 90')  # VAT CUI/CIF (RO prefix is stripped)
+'18547290'
 >>> validate('185 472 90')  # non-VAT CUI/CIF
 '18547290'
 >>> validate('1630615123457')  # CNP
@@ -37,8 +37,12 @@ from stdnum.util import clean
 
 def compact(number: str) -> str:
     """Convert the number to the minimal representation. This strips the
-    number of any valid separators and removes surrounding whitespace."""
-    return clean(number, ' -').upper().strip()
+    number of any valid separators and removes surrounding whitespace. The
+    optional RO country prefix is removed to return the domestic form."""
+    number = clean(number, ' -').upper().strip()
+    if number.startswith('RO'):
+        number = number[2:]
+    return number
 
 
 # for backwards compatibility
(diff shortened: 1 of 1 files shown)


arthurdejong:
Hi @gaoflow,

This change fails the tests. In particular this number is accepted as valid while it is not: `RO RO 16621241`

Also see the discussion in #485 that describes some of the history and complexity of the current implementation.

binggao1230:
Thanks for catching that. Fixed in GPG-verified `87f99de`.

The issue was that after `cf.compact()` stripped one optional `RO`, the remaining `RO16621241` still went through `cui.validate()`, whose own compacting accepts `RO` again. I added an explicit domestic-form digit check in `cf.validate()` after its single prefix strip, so `RO 185 472 90` still returns `18547290`, while `RO RO 16621241` now raises `InvalidFormat`.

Verification:

```text
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest -o addopts='' --doctest-modules --doctest-glob='*.doctest' tests/test_eu_vat.doctest stdnum/ro/cf.py -q
2 passed
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest -o addopts='' --doctest-modules stdnum/ro -q
4 passed
```

PR #501, opened 2026-07-07
Update Taiwan UBN check digit for the 2023 rule change (divisible by 5)

gccbb02498:
## Summary

Taiwan's Ministry of Finance relaxed the check digit logic for the UBN
(Unified Business Number, 統一編號) effective **2023-04-01**. The weighted sum
now only needs to be **divisible by 5** instead of 10, enlarging the pool of
assignable numbers as the old space is being exhausted.

- Numbers valid under the old rule (divisible by 10) remain valid under the
  new one, since divisibility by 10 implies divisibility by 5. No previously
  valid number is rejected — consistent with the guideline that marking valid
  numbers as invalid must be avoided.
- The special case for a `7` in the 7th position (where 1 may be added to the
  sum before the divisibility check) is preserved, now evaluated against
  divisibility by 5.

## Changes

- `stdnum/tw/ubn.py`: `calc_checksum` now returns the weighted sum mod 5;
  `validate` accepts `checksum == 0`, or `checksum == 4` when the 7th digit is
  `7`. The module docstring documents the rule change with an official reference.
- `tests/test_tw_ubn.doctest`: added old-rule numbers (divisible by 10,
  including the 7th-digit-is-`7` case) to confirm they stay valid, plus new-rule
  numbers (divisible by 5 but not 10, including the 7th-digit-is-`7` case).

## Reference

Fiscal Information Agency, Ministry of Finance — explanation of the revised
UBN check digit logic:
https://www.fia.gov.tw/singlehtml/3?cntId=c4d9cff38c8642ef8872774ee9987283

### DIFF
--- a/stdnum/tw/ubn.py
+++ b/stdnum/tw/ubn.py
@@ -22,9 +22,15 @@ The Unified Business Number (UBN, 統一編號) is the number assigned to busine
 within Taiwan for tax (VAT) purposes. The number consists of 8 digits, the
 last being a check digit.
 
+Since 2023-04-01 (民國112年) the Ministry of Finance relaxed the check digit
+logic so that the weighted sum only needs to be divisible by 5 (instead of 10)
+to increase the pool of available numbers. Numbers that were valid under the
+old rule (divisible by 10) remain valid under the new one.
+
 More information:
 
 * https://zh.wikipedia.org/wiki/統一編號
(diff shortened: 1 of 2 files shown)


arthurdejong:
Hi @gccbb02498,

Thanks for providing this fix. It is merged as 5fd7498.

PR #503, opened 2026-07-17
Add support for ISO 4914 UPI validation

JonArnfred:
## Summary

Add support for validating ISO 4914 Unique Product Identifiers (UPIs).

A UPI is a 12-character financial identifier for over-the-counter derivative products. The implementation validates:

- the mandatory `QZ` prefix;
- the restricted alphanumeric alphabet;
- the 12-character length; and
- the ISO 7064 Mod 31, 30 check character.

It also provides the standard `compact()`, `validate()`, `is_valid()`, and `calc_check_digit()` interfaces.

## References

- [ISO 4914](https://www.iso.org/standard/80506.html)
- [Derivatives Service Bureau UPI information](https://www.anna-dsb.com/upi-/)

## Testing

- Added module documentation examples.
- Added a dedicated doctest covering normalization, prefix, alphabet, length, and checksum validation.
- Included four published valid UPIs sourced from the DSB, CME Group, and official regulatory material.
- The new module achieves 100% test coverage.
- Full test suite: 407 passed, 9 skipped.
- Flake8 and strict mypy checks pass.

Publicly documented real UPI examples appear to be limited, so the test set currently contains four published examples. The official DSB archive requires registration, I can look into that if we need more examples.

### DIFF
--- a/tests/test_upi.doctest
+++ b/tests/test_upi.doctest
@@ -1,0 +1,82 @@
+test_upi.doctest - more detailed doctests for the stdnum.upi module
+
+
+This library is free software; you can redistribute it and/or
+modify it under the terms of the GNU Lesser General Public
+License as published by the Free Software Foundation; either
+version 2.1 of the License, or (at your option) any later version.
+
+This library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


tests/test_upi.doctest
```diff
@@ -0,0 +1,82 @@
+test_upi.doctest - more detailed doctests for the stdnum.upi module
+
+
```
arthurdejong, line 3: This file is missing a copyright statement.

arthurdejong:
Hi @JonArnfred,

Thanks for your contribution. I've merged it as 2413a52.

Kind regards

tests/test_upi.doctest
JonArnfred: Good catch!

JonArnfred:
Thanks for merging.

Issue #450, opened 2024-08-02
Add support for Czech/Slovak IČ - company registration number


PavelPancocha:
Just 8 digits.

Example: https://or.justice.cz/ias/ui/rejstrik-firma.vysledky?subjektId=47718&typ=PLATNY&sp=H4sIAAAAAAAAAKtWykvMTU0JSCwCUiWpRcVKVtHVStmplUpWQJmq1DIlHaWyxJzSVCD%252F6MLs%252FJRE%250D%250AhcTSknyl2thaAOBvbwI7AAAA%250D%250A

Example: 00177041, could be formatted as `001 77 041`

dbartenstein:
@PavelPancocha is there any official API in the Czech Republic where the company registration number could be validated? See example for Germany: https://github.com/arthurdejong/python-stdnum/blob/6cbb9bc09c25fbda7a032521bc57b44e0ce18ec4/stdnum/de/handelsregisternummer.py#L328

arthurdejong:
Is there a relation to the DIČ number (which is included in python-stdnum already)?

Btw, see https://github.com/arthurdejong/python-stdnum/blob/master/CONTRIBUTING.md#adding-number-formats for what information would be useful when adding a new number format.

PavelPancocha:
In most cases the IČ(O) (its the same thing) is just the DIČ without CZ prefix, but can be different - mainly for physical persons. Also if you are now VAT payer, you don't provide the VAT on invoices, it is only used internally with tax-office.

You can use this official API to get info about specific number: https://ares.gov.cz/swagger-ui/#/ekonomicke-subjekty-szr/vratEkonomickySubjektSzr

DMZ22:
Opened #504 for this — adds `stdnum.cz.ico` and `stdnum.sk.ico` (8-digit IČO with the modulo-11 check digit). To @arthurdejong's question: for a legal entity the IČO is the same number as the Czech DIČ without the `CZ` prefix, so the Czech module reuses the check-digit algorithm already in `cz/dic`, and the Slovak module delegates to the Czech one (mirroring `sk/rc` → `cz/rc`). Included 21 real Czech and 21 real Slovak numbers as regression tests.

PR #504, opened 2026-07-23
Add Czech and Slovak IČO validation

DMZ22:
This implements #450: validation for the **IČO**, the organisation identifier used in the Czech Republic and Slovakia.

### Format
- **Name:** IČO — *Identifikační číslo osoby* (CZ) / *Identifikačné číslo organizácie* (SK), also written IČ.
- **Structure:** 8 digits, the last being a weighted modulo-11 check digit (weights 8…2 over the first 7 digits, check digit `(11 - sum) % 11`, mapped so a remainder of 0 → `1`).
- Identifies a legal entity or sole trader; listed in the Czech ARES register and the Slovak ORSR / RPO registers.

### Relation to DIČ (raised in the issue)
For a legal entity the IČO is exactly the digits of the Czech VAT number (DIČ) without the `CZ` prefix, so `cz/ico` uses the same check-digit computation that `cz/dic` already applies to its 8-digit legal-entity form. The Slovak IČO is identical to the Czech one (until 1993 they were a single country), so `sk/ico` delegates to `cz/ico`, mirroring the existing `sk/rc` → `cz/rc` pattern.

### Sources
- https://cs.wikipedia.org/wiki/Identifikační_číslo_osoby
- Official register / validation service: https://ares.gov.cz/ (CZ), https://www.orsr.sk/ (SK)

### Tests
- Module docstrings cover valid input, invalid check digit, wrong length and non-digit input.
- `tests/test_cz_ico.doctest` and `tests/test_sk_ico.doctest` each bulk-validate 21 real numbers taken from public company records (e.g. Škoda Auto `00177041`, ČEZ `45274649`; Slovnaft `31322832`, Tatra banka `00686930`).
- All doctests pass under `NORMALIZE_WHITESPACE`/`IGNORE_EXCEPTION_DETAIL`; every line of the new modules is exercised (100% coverage).

Happy to add a `format()` grouping or an `IČO → DIČ` helper if you'd find either useful.

### DIFF
--- a/stdnum/cz/ico.py
+++ b/stdnum/cz/ico.py
@@ -1,0 +1,92 @@
+# ico.py - functions for handling Czech organisation identifiers
+# coding: utf-8
+#
+# Copyright (C) 2026 Devashish Moghe
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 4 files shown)


arthurdejong:
Hi @DMZ22,

Thanks for providing this implementation. It was merged as b0ce222.

Issue #408, opened 2023-03-08
 Add support for Oman TIN


unho:
* https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/Oman-TIN.pdf
  * Either 7 digits, or 12 alphanumeric characters.
    * My suspicion is that the 12 alphanumeric are `OM` then followed by 10 digits.

Online validation:
* Using the `VATIN` on https://www.taxoman.gov.om/portal/web/taxportal/tax-data-validation

Examples:
* `OM1100006083` from https://www.omantel.om/wcm/connect/7041d8b4-9865-4974-9980-97c13a7589e6/VatCertificate.pdf?MOD=AJPERES&CVID=nzr8s9j&CVID=nzr8s9j&CVID=nzr8s9j
* https://www.google.com/search?q=site%253A.om+%22VATIN%22
* https://www.google.com/search?q=site%253A.om+%22TIN%22

DMZ22:
Opened #505 adding `stdnum.om.vat` for the Oman VATIN (the `OM` + 10-digit VAT/Excise form — 12 chars, no check digit, validated structurally like `gh/tin.py`). Left out the 7-digit entity "tax card number" since it's an unstructured all-digit sequence; happy to add it or fold both into `om/tin.py` if you'd prefer.

PR #505, opened 2026-07-24
Add Oman VAT number (VATIN) validation

DMZ22:
Partially addresses #408 by adding the Oman **VATIN** — the VAT/Excise form of the Oman TIN, which is the one with a defined structure.

### Format
- **Name:** VATIN (VAT identification number), issued by the Oman Tax Authority.
- **Structure:** the letters `OM` followed by 10 digits — 12 characters in total.
- **No check digit.** Numbers are verified online via the tax authority portal, so this module validates length and formatting only (in the spirit of e.g. `stdnum/gh/tin.py`, which is also structural).

### Sources / validation service
- https://tms.taxoman.gov.om/portal/web/taxportal/vatin-validation (official VATIN validation)
- OECD CRS TIN guidance for Oman describes the 12-character VAT/Excise TIN.

### Tests
- Module docstring covers valid input, whitespace/lowercase normalisation, missing prefix, wrong length and non-digit tail.
- `tests/test_om_vat.doctest` bulk-validates real numbers taken from published Omani VAT certificates (e.g. `OM1100006083`, `OM1100018164`, `OM1100005523`). Every line of the module is exercised.

### Note
The `#408` issue also mentions a 7-digit "tax card number" for entities. I left that out: it's an unstructured all-digit sequence with no fixed length constraint or check digit, so validating it adds little (per CONTRIBUTING's guidance on numbers whose only spec is "N digits"). Happy to add it, or to fold both into an `om/tin.py`, if you'd prefer.

### DIFF
--- a/stdnum/om/__init__.py
+++ b/stdnum/om/__init__.py
@@ -1,0 +1,19 @@
+# __init__.py - collection of Omani numbers
+# coding: utf-8
+#
+# Copyright (C) 2026 Devashish Moghe
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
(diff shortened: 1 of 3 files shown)


arthurdejong:
Hi @DMZ22,

Thanks for contributing this. The changes were merged as 9f1bc4a. The only significant change I made was to ensure that numbers that end with X are also supported (they were found "in the wild" and confirmed to be valid via the validation service).

That triggered me to reverse engineer the check digit algorithm in 23be163. I did not manage to find weights for the first 4 digits (after OM) as there was insufficient variation there but this algorithm works for all the valid numbers I managed to find.

PR #508, opened 2026-08-12
Strip comma instead of keyboard-adjacent m when compacting CID

holvitolv:
It does look like once upon a 2014 the author of stdnum.eu.at_02 hit `m` when intending to hit `,` while typing the list of characters to strip from CIDs, as the [documentation](https://www.europeanpaymentscouncil.eu/sites/default/files/kb/file/2025-10/EPC262-08%20v12.0%20Creditor%20Identifier%20Overview.pdf) for it offers no more plausible theories that I can tell, no special meaning is given to an `m` or `M` or their complement there :D

E.g. the Maltese example MT50ZZZ670169305T from the documentation, if provided in lowercase, would lose its M in compacting, as demonstrated by the test in the first commit. Strip the keyboard-adjacent punctuation instead.


arthurdejong:
Hi @holvitolv,

Thanks for fixing this. It has been merged as 30aaa38.

